PR #10, opened 2015-06-23
Display Speed Calc results

Kiru4:
Speed total is modified by choice scarf and coloured in blue (to mark an item modifying speed total).
FIRST is displayed if the pokemon attacking is faster than the other one.
LAST is displayed if the pokemon attacking is slower (or equal : speed tie)  than the other one.


Zarel:
Why is this diff unreadable? What did you do? Can you rebase on the latest master and squash?

PR #11, opened 2015-06-24
Display Speed Calc results

Kiru4:
Speed total is modified by choice scarf and coloured in blue (to mark an item modifying speed total).
FIRST is displayed if the pokemon attacking is faster than the other one.
LAST is displayed if the pokemon attacking is slower (or equal : speed tie) than the other one.

This time you should properly see the diffs.

### DIFF
--- a/index.html
+++ b/index.html
@@ -7,6 +7,12 @@
     <link type="text/css" rel="stylesheet" href="./select2/select2.css" />
     <link type="text/css" rel="stylesheet" href="./ap_calc.css?v2" />
     <link type="text/css" rel="stylesheet" href="./bootstrap.css" />
+	
+	<style type="text/css">
+		.total{
+			color:black;	
+		}
+	</style>
 </head>
 
 <body>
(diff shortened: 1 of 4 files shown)


AustinXII:
@Kiru4 do you have any intent on squashing?

Kiru4:
I would like the modifications to be integreted into the trunk. Have submitted the commit a while ago...

Zarel:
Your code doesn't fit our code style, and I really don't understand what this is supposed to change. Can you describe the change?

Kiru4:
I did not change your "code style" added function where you added yours etc.
The change allow the user to know which pokemon is faster than the other which is really important.
Plus a pokemon taking an item that changes the value of its statistics display the new statistics values and their display are modified to let the user know the changes.

Zarel:
I'm talking about how you do things like

    if ( STATS[i] == "sp" )
    {

but our code style would be

    if (STATS[i] == 'sp') {

AustinXII:
@Kiru4 the speed is already displayed in battle on PS, I think this PR is kind of pointless

Kiru4:
If the "{" is not placed properly you could easily change it.
@AustinXII I have done it for scarf user and for people building their team to be faster than specifics mons and for people playing on console.

Zarel:
Having looked at this more closely, it doesn't seem like it implements this very efficiently. The UI also seems kind of crude. There's a spot to the right of the stat boost that could be used to represent boosted stats.

Since it would need to be rewritten from scratch before I'd accept it anyway, I'm closing this.

Kiru4:
Ok, hopefully you will do it properly (taking care of the boost + and -), just wanted to help and add features that I needed.

PR #12, opened 2015-07-09
RBY movesets update

Guernouille:
Adding missing movesets that are relevant in their respective tiers


Zarel:
@TheImmortal or @Joimer, is this acceptable?

Joimer:
Yes, the new movesets are used.

PR #15, opened 2015-09-01
Update damage.js

jitox:
Add damage calculation when using Venoshock and foe is poisoned


Zarel:
Oh, it looks like you have, nevermind on my comment from your other pullreq.

PR #16, opened 2015-09-01
Update move_data.js

jitox:
Venoshock does not appear as a move on the calculator. Besides this change it has to hit double when foe is poisoned


Zarel:
I wonder if it's possible to support the doubling...

PR #20, opened 2015-09-04
Add updated Mega sets, Add Hoopa + hoopa unbound and Update Various o…

AustinXII:
…ther sets

Added Mega Latias + Glalie aswell, and updated Beehyeem and Archeops


Zarel:
Wow, this is way too unminified. Maybe something like one line per set would work better.

PR #30, opened 2015-09-29
Add Multiple Moves and Fix some damage Calcs

AustinXII:


AustinXII:
Will Squash later, will also probably add more moves, Please do not merge.

AustinXII:
reminder: Future Sight

AustinXII:
@Marty-D  Please review

js/data/item_data.js
```diff
@@ -354,3 +360,11 @@ function getNaturalGift(item) {
     }
     return {'t':'Normal','p':1};
 }
+function getTechnoBlast(item) {
```
Marty-D: Seems like this should be done the same way as `getItemBoostType(item)`, just returning the appropriate type.

js/damage.js
```diff
@@ -305,7 +308,10 @@ function getDamageResult(attacker, defender, move, field) {
         description.weather = field.weather;
     } else if (gen >= 6 && move.name === "Knock Off" && !(defender.item === "" ||
             (defender.name === "Giratina-O" && defender.item === "Griseous Orb") ||
-            (defender.name.indexOf("Arceus") !== -1 && defender.item.indexOf("Plate") !== -1))) {
+            (defender.name.indexOf("Arceus") !== -1 && defender.item.indexOf("Plate") !== -1) ||
+            (defender.name.indexOf("Genesect") !== -1 && defender.item.indexOf("Drive") !== -1) ||
+            (defender.name.indexOf("Primal Groudon") && defender.item === "Red Orb") ||
+            (defender.name.indexOf("Primal Kyogre") && defender.item === "Blue Orb"))) {
```
Marty-D: Just checking for `"Groudon"` and `"Kyogre"` here is fine (like the ones above), since this stuff applies to all formes of these Pokemon.

Marty-D:
Looks good other than those few changes to make :+1:

js/data/item_data.js
AustinXII: @Marty-D Oddly that was my thought process aswell but for what ever reason I went the other way with it

AustinXII:
@Marty-D finished

Zarel:
Notice your commit summary.

js/damage.js
```diff
@@ -86,6 +86,8 @@ function getDamageResult(attacker, defender, move, field) {
         description.moveType = move.type;
     } else if (move.name === "Judgment" && attacker.item.indexOf("Plate") !== -1) {
         move.type = getItemBoostType(attacker.item);
+    } else if (move.name === "Techno Blast" && attacker.item.indexOf("Drive") !== -1) {
+        move.type = getItemBoostType(attacker.item);
```
Marty-D: This isn't really what I meant; you can keep your other function but just implement it like this function. Doing it in the same function will cause Drives to boost same-type attacks like Plates and the rest of them. [For reference](https://github.com/Zarel/honko-damagecalc/pull/30/files#diff-63ddc2b0f3b50a57a4f0b5e52d659438R281).

js/damage.js
```diff
@@ -305,7 +307,10 @@ function getDamageResult(attacker, defender, move, field) {
         description.weather = field.weather;
     } else if (gen >= 6 && move.name === "Knock Off" && !(defender.item === "" ||
             (defender.name === "Giratina-O" && defender.item === "Griseous Orb") ||
-            (defender.name.indexOf("Arceus") !== -1 && defender.item.indexOf("Plate") !== -1))) {
+            (defender.name.indexOf("Arceus") !== -1 && defender.item.indexOf("Plate") !== -1) ||
+            (defender.name.indexOf("Genesect") !== -1 && defender.item.indexOf("Drive") !== -1) ||
+            (defender.name.indexOf("Groudon") && defender.item === "Red Orb") ||
+            (defender.name.indexOf("Kyogre") && defender.item === "Blue Orb"))) {
```
Marty-D: You still have to check the `indexOf()` value like the others, otherwise this will work for every Pokemon _except_ Groudon and Kyogre.

AustinXII:
@Marty-D  Ok tested everything and It works, I think we should be good.

Zarel:
I said, pay attention to your commit summary........

Zarel:
image

AustinXII:
Sorry @Zarel I did this at school and didn't see your comment

Marty-D:
Now that I've had time to think more about it, implementing the Blue/Red Orb in the calc is kind of a waste. You'll notice that none of the Mega Stones are implemented either, and that's because having no item on a Mega Pokemon is equivalent to having the appropriate Mega Stone on it for the purposes of Knock Off.

The only reason the Griseous Orb, Plates, and Drives are useful is because they do something other than just preventing Knock Off from doing more damage on the right Pokemon.

AustinXII:
ok finished @Marty-D I left the items in on the setdex just in case we need it in the future

PR #33, opened 2015-10-13
Add New Pokemon

AustinXII:
**DO NOT MERGE YET** 
Added Zygarde forms since they were officially announced on the pokemon website, will update as more information is given.
edit: I will be using this Pull request for all the new pokemon released in the new games whenever that is

### DIFF
--- a/js/data/pokedex.js
+++ b/js/data/pokedex.js
@@ -10579,5 +10579,220 @@ var POKEDEX_XY = $.extend(true, {}, POKEDEX_BW, {
     },
     "w": 305.0,
     "ab": "Aura Break"
+  },
+  "Zygarde-Cell": {
+    "t1": "Dragon",
+    "t2": "Ground",
+    "bs": {
+      "hp": ,
+      "at": ,
+      "df": ,
+      "sa": ,
+      "sd": ,
+      "sp": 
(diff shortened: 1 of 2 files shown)


js/data/pokedex.js
```diff
@@ -10579,5 +10579,75 @@ var POKEDEX_XY = $.extend(true, {}, POKEDEX_BW, {
     },
     "w": 305.0,
     "ab": "Aura Break"
+  },
+  "Zygarde-Cell": {
+    "t1": "Dragon",
+    "t2": "Ground",
+    "bs": {
+      "hp": ,
```
Zarel: You can use placeholder values like `0` or `undefined` or something, instead of invalid JavaScript.

AustinXII: I left it blank so I could just add the actual info when its released, this shouldn't be merged any time anyways but I could fix it if you prefer for some reason?

Zarel: It's no big deal, so no need to fix it now, it's just for future reference. Invalid code just makes me vaguely uncomfortable. :P

AustinXII: @Zarel  yea its understandable lol

PR #40, opened 2015-10-22
Add missing Gen 5 moves and fix other moves

AustinXII:


js/data/move_data.js
```diff
@@ -1430,11 +1436,6 @@ var MOVES_BW = $.extend(true, {}, MOVES_DPP, {
         category: 'Physical',
         makesContact: true
     },
-    'Attack Order': {
```
AustinXII: duplicate

js/damage.js
```diff
@@ -305,6 +305,8 @@ function getDamageResult(attacker, defender, move, field) {
         bpMods.push(0x800);
         description.moveBP = move.bp / 2;
         description.weather = field.weather;
+    } else if (defender.weight <= "200" && move.name === "Sky Drop" || move.name === "Sky Drop" && (defender.type1 === "Flying" || defender.type2 === "Flying")) {
```
Marty-D: Several issues here, not least of which is this doesn't work at all.
- This should be done in `getDamageResult()` instead
- 200 shouldn't be a string
- You should be checking if the weight is >=, not less than

Basically `(move.name === "Sky Drop" && (defender.weight >= 200 || (defender.type1 === "Flying" || defender.type2 === "Flying")))` would be a better check.

js/data/move_data.js
```diff
@@ -1494,6 +1502,13 @@ var MOVES_BW = $.extend(true, {}, MOVES_DPP, {
         category: 'Special',
         isBullet: true
     },
+    'Electro Web': {
```
Marty-D: Electroweb is one word.

AustinXII:
@Marty-D updated

js/damage.js
```diff
@@ -120,6 +120,9 @@ function getDamageResult(attacker, defender, move, field) {
     if (typeEffectiveness === 0) {
         return {"damage":[0], "description":buildDescription(description)};
     }
+    if (move.name === "Sky Drop" && (defender.weight >= 200 || (defender.type1 === "Flying" || defender.type2 === "Flying"))) {
```
Marty-D: I forgot this file is also used for Gen 5 damage calcs, so this should actually be something like `(move.name === "Sky Drop" && (defender.type1 === "Flying" || defender.type2 === "Flying" || (gen >= 6 && defender.weight >= 200)))`.

AustinXII:
done @Marty-D

Marty-D:
:+1:

PR #42, opened 2015-10-30
various set updates

AustinXII:


js/data/setdex_xy.js
```diff
@@ -15,7 +15,17 @@ var SETDEX_XY = {
 	"Accelgor": {
 		"RU Spikes Lead": {"level":100,"evs":{"hp":252,"sp":252,"sa":4},"nature":"Timid","ability":"Sticky Hold","item":"Focus Sash","moves":["Spikes","Final Gambit","Encore","Bug Buzz"]}
 	},
-	"Aegislash": {
+	"Aegislash-Blade": {
```
Zarel: Why are these separate?

AustinXII: they are two different forms with two different set of stats.

PR #43, opened 2015-11-02
Add Synchronoise and fix who should be contacted in case of an error

AustinXII:


js/damage.js
```diff
@@ -120,9 +120,13 @@ function getDamageResult(attacker, defender, move, field) {
     if (typeEffectiveness === 0) {
         return {"damage":[0], "description":buildDescription(description)};
     }
-    if (move.name === "Sky Drop" && (defender.type1 === "Flying" || defender.type2 === "Flying" || (gen >= 6 && defender.weight >= 200))) {
+    if (move.name === "Sky Drop" && (defender.type1 === "Flying" || defender.type2 === "Flying" || (gen >= 6 && defender.weight >= 200) || field.isGravity)) {
         return {"damage":[0], "description":buildDescription(description)};
     }
+    if (move.name === "Synchronoise" &&
```
AustinXII: hmm I wish I didn't do this so late, I believe http://hastebin.com/qugizivayo.avrasm was the correct way to do this, can you confirm @Marty-D ?

Marty-D: No, what you did is fine. That would only check `type1` against `type1` since it always exists.

What I'm wondering is why the Sky Drop change? None of the other moves disabled by Gravity are covered either. If we want to do that, might as well flag them in the movedex and check it only once somewhere else.

Marty-D: Actually, what you did isn't fine, specifically because `type2` might be blank. If both `type2` are blank then it doesn't pass. I'll fix it later.

AustinXII: the sky drop change was only added because I noticed I missed it when making the Synchronoise addition, I didn't think of adding more as I thought they would have already been covered.

PR #49, opened 2015-12-15
Fix Venoshock 4x effect on poison

luiscarloscb:
Currently Venoshock calculates 4x on "Poisoned" status due to the bonus damage calculation ocurring twice, once in the lines removed from the commit and the second time in [this check](https://github.com/Zarel/honko-damagecalc/blob/master/js/damage.js#L308).

Also the removed lines had the issue of ignoring "Badly Poisoned" status.


Marty-D:
Thanks! :+1:

PR #56, opened 2016-03-23
Update XY setdex.

Honko:
Finally got around to writing something that could generate this from
the new smogon api. I'll try to run it monthly-ish from now on.


Honko:
Just checking: are you planning on updating the server with this and just haven't had a chance to yet, or are you expecting me to? I'm fine either way, but I just noticed it wasn't updated yet and tried scping it over like I used to and got "Permission denied". So if you're expecting me to do it, I'll need permission to write to that directory again.

Zarel:
I can.

PR #60, opened 2016-05-18
Allow the use of Soul dew on Mega Latias / Mega Latios

AustinXII:
Relevant for OMs


Zarel:
Maybe you should mark that somehow. "Illegal combination" or "Hackmons only" or something to make it clear it only works on Hackmons-relevant formats.

AustinXII:
@Zarel Updated

Zarel:
Your update doesn't seem to have addressed my concern?

AustinXII:
Note: Won't be updated for a month

PR #64, opened 2016-06-16
RBY: Some NFE used in UU + minor fixes

Guernouille:
Adding Graveler, Haunter and Kadabra in UU, putting back Thunder Tauros for OU, changing Sand Attack for Rock Slide on Dugtrio as Articuno and Moltres are relevant in UU.


Honko:
These sets look fine, but they'll just get overwritten next time I run the script to import sets from the Smogon dex. I don't maintain these files by hand. Can you work with whoever's in charge of RBY's Smogon analyses (maybe Jellicent?) to get these sets uploaded to the dex?

Zarel:
What @Honko said.

Guernouille:
Yup, I'll see that on Smogon then :P

PR #75, opened 2016-11-18
Add missing pokemon and fix base stats for S/M and Add Z-Moves

AustinXII:


AustinXII:
@Zarel @Marty you can merge this now

PR #79, opened 2016-11-27
Fix 'Long Reach' and 'Parental Bond' calculations

dramamine:
Was reviewing the SuMo update and found Ability was capitalized when it shouldn't be. Parental Bond wasn't broken but this is more concise; I tested this one in browser to make sure it's working properly.

### DIFF
--- a/js/damage.js
+++ b/js/damage.js
@@ -285,7 +285,7 @@ function getDamageResult(attacker, defender, move, field) {
     } else if (defAbility === "Fluffy" && move.makesContact && move.type === "Fire") {
         bpMods.push(0x2000); 
         description.defenderAbility = defAbility;
-    } else if (defAbility === "Fluffy" && (move.makesContact && !attacker.Ability === "Long Reach")) {
+    } else if (defAbility === "Fluffy" && move.makesContact && !attacker.ability === "Long Reach") {
         bpMods.push(0x800);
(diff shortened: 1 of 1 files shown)


js/damage.js
```diff
@@ -342,7 +342,7 @@ function getDamageResult(attacker, defender, move, field) {
             (attacker.ability === "Strong Jaw" && move.isBite)) {
         bpMods.push(0x1800);
         description.attackerAbility = attacker.ability;
-    } else if (attacker.ability === "Tough Claws" && (move.makesContact && attacker.Ability !== "Long Reach")) {
+    } else if (attacker.ability === "Tough Claws" && (move.makesContact && attacker.ability !== "Long Reach")) {
```
Honko: Not your mistake, but the check for Long Reach is unneeded here. If the attacker's ability is Tough Claws it can't also be Long Reach.

AustinXII: my mistake, you can also remove it from the fluffy check as that also doesn't make any sense

dramamine:
OK, updated that Tough Claws line per your comments.

AustinXII:
@Zarel @MartyD

js/damage.js
AustinXII: Never mind I'm clearly misreading..

PR #81, opened 2016-12-04
Implement new Water Shuriken mechanics

AustinXII:


Zarel:
Is it always 20, or only 20 after getting a KO?

AustinXII:
It is always 20 if the greninja is ash greninja, speaking of it, now looking at ash-greninja it should definitely be its own Pokemon; will do this shortly

Zarel:
You normally can't get an Ash-Greninja without getting a KO, is what I'm saying.

Don't confuse Battle Bond Greninja with Ash-Greninja. :|

AustinXII:
@Zarel I really don't think that matters for calculation purposes... who cares what you need to do in order to get a Pokémon people only care about the damage outputs. Anyways I probably was getting them confused (thought they were basically the same thing) but I don't think it matters unless there's a huge difference that I'm missing/not seeing.

AustinXII:
Even if there is it won't matter once Greninja-Ash is it's own Pokemon because from what I'm seeing 
Battle Bond === Regular just different abilities  while Ash (once it's transformed into this) has increased water shuriken damage and stats

Zarel:
The huge difference is, Battle Bond Greninja does not get the boosted Water Shuriken damage.

AustinXII:
@Zarel Yes I don't believe what I said is opposing that

AustinXII:
@Zarel updated

PR #84, opened 2016-12-06
Fixed a missing mechanism of "Water Bubble" ability

Czestchan:
#84 Implement: Added water-type dmg boost of ability "Water Bubble"


Marty-D:
Water Bubble is almost certainly implemented as a base power modifier rather than a final modifier. I notice Steelworker is also down here in what is likely the wrong place.

Zarel:
Should probably merge this now and investigate a proper implementation later.

PR #85, opened 2016-12-09
Update Aerilate-like abilities

MagiaSN:
According to Serebii.net and several sites, the power boost of Aerilate, Pixilate, Refrigerate and Galvanize has been decreased from 30% to 20%.

### DIFF
--- a/js/damage.js
+++ b/js/damage.js
@@ -343,7 +343,7 @@ function getDamageResult(attacker, defender, move, field) {
     }
     
     if (isAerilate || isPixilate || isRefrigerate || isGalvanize) {
-        bpMods.push(0x14CD);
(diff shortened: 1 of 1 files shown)


AustinXII:
If you're doing it like this you need to have a generation check, the reason I held off on the ate changes is because I was hoping to have gotten time to do the Z moves already :|

MagiaSN:
Oh, I didn't think about that. When I saw you merged pull requests about Water Bubble and so on I thought you are only focusing on gen 7. Now that you have noticed this issue, you may close it if you like. Wish you all the best!

MagiaSN:
Already updated. Hope it helps.

PR #87, opened 2016-12-14
Q tesseract patch 2

KrisOnPS:


Zarel:
???

AustinXII:
If this is for OM's or something, we don't support it on the calc, if you really need to calc something just change the typing/stats manually on the calc.

PR #94, opened 2016-12-16
Add autoset terrain feature

MagiaSN:
Autoset terrain if a surge ability is selected.
Also fixed a small bug in terrain transition: status list won't be enabled if change Misty Terrain to Electric Terrain.


AustinXII:
@MagiaSN what's your ps username?

MagiaSN:
MagiaSN :)

AustinXII:
:+1:

PR #96, opened 2016-12-22
Fix: remove old buildDescription function

MagiaSN:
There are two `buildDescription` function, one defined in damage.js and the other defined in damage_rse.js. Now the one in damage_rse.js would be called, which cause terrain message not shown in the result. I think there is no need to keep the one in damage_rse.js because the only difference between them are terrain and Friend Guard.


AustinXII:
Did you test this? I swear it didn't work when I removed this function in testing

MagiaSN:
@AustinXII Yes, I have tested this. 
Before I remove this function, when I use the default Abomasnow set to calculate and change terrain to Grassy Terrain, the result is shown as "252 SpA Life Orb Abomasnow Giga Drain vs. 0 HP / 0 SpD Abomasnow: 109-129 (33.9 - 40.1%) ...", in which the terrain message is missing. When I put a breakpoint at the end of function `getDamageResult` (defined in damage.js), I find `buildDescription` in damage_rse.js is called.
After I remove this function, when I use the same test case, the result is shown as "252 SpA Life Orb Abomasnow Giga Drain vs. 0 HP / 0 SpD Abomasnow **in Grassy Terrain**: 109-129 (33.9 - 40.1%)", in which the terrain message is correctly shown. When I put the same breakpoint again, `buildDescription` in damage.js is called.
I think there won't be any problem if I remove this, because there is still a `buildDescription` in damage.js. By the way my browser is Microsoft Edge if you want to know.

AustinXII:
Terrains were introduced in gen 6

AustinXII:
Ugh this was wrong.. the correct way would be to add the terrain check to this function and remove the other function in damage.js since this removed detailed descriptions from old gens :/

Zarel:
Feel free to revert. Do you need help with the reverting process?

AustinXII:
@Zarel nah i think I know how

MagiaSN:
@AustinXII Sorry if I have brought you trouble, but I have compared the one in damage.js and this, and found the only difference is that the previous one has terrain check and Friend Guard check, so I think it's safe to delete this one. Did I miss something?

AustinXII:
@MagiaSN yes if you remove the builddescription function from damage_rse.js instead of damage.js ADV And DPP will have very limited damage output descriptions

AustinXII:
@MagiaSN also don't worry about it

MagiaSN:
@AustinXII I have tested ADV and DPP and it seems to work fine. Could you show me when this will happen? :)

PR #101, opened 2017-01-14
Improve z-move input

MagiaSN:
Add a "z" checkbox beside move selector. If it is checked, the corresponding z-move will be used for calc, which should simplify z-move input.

This requires work of @AustinXII (in progress), which adds a z-power property to each move, and a z-move map.

Changes in detail:

1. Add a "z" checkbox beside each move (`index.html`, `calc_bc.html`), and update layout to make all views in the same line. (`ap_calc.css`)
2. Add `getZMoveName` function, which finds the proper z-move to use. Hidden Power becomes Breakneck Blitz. Special move with special z-crystal becomes special z-move (e.g. Volt Tackle with Pikanium Z becomes Catastropika). Otherwise a generic z-move is used. Generic z-moves do not force require a z-crystal or it would be too difficult to use. (`ap_calc.js`)
3. Add z-move check to `getMoveDetails`. In gen 7, when z checkbox is checked and the original has a `zp` property, we use z-move instead of original move. (`ap_calc.js`).
4. Uncheck z checkbox if move was changed. (`ap_calc.js`)
5. Show power of generic z-moves in damage description. (`damage.js`)

### DIFF
--- a/ap_calc.css
+++ b/ap_calc.css
@@ -324,20 +324,29 @@ select.toxic-counter {
     font-size: 0.9em;
 }
 .move-selector {
-    width: 9.5em;
+    width: 8.5em;
 }
 .poke-info input.move-bp {
     width: 2em;
+}
+.move-type {
+    width: 5.5em;
 }
 .move-cat {
     width: 5.7em;
 }
 .move-hits {
-    width: 5em;
+    width: 4.5em;
 }
 .poke-info .crit-btn {
     font-size: 0.8em;
     width: 2.5em;
+    height: 1em;
+    padding: 3px 3px;
+}
+.poke-info .z-btn {
+    font-size: 0.8em;
+    width: 1.5em;
     height: 1em;
     padding: 3px 3px;
 }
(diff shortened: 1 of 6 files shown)


AustinXII:
Awesome work!

Zarel:
Someone tell me which of these I should be merging (or both?)

AustinXII:
@Zarel I'll merge it when it's ready

AustinXII:
I still haven't had enough time to finish, but I added all the work that I had done, there's still a lot to go through and add though D: Someone can feel free to finish the remaining Z-moves.  @MagiaSN you should also test this, I don't know if you based your work of what I had finished when I suggested you look at the work that I had done.

MagiaSN:
@AustinXII I may finish it in a day or two. By the way what's you way of testing? I didn't see any test code in the repo, so I always test manually, by opening the calculator in a browser to see if it works as expected. Is that enough?

AustinXII:
Yeah that's enough

MagiaSN:
@AustinXII @Zarel You can merge now

AustinXII:
@Zarel I'll test this in a few hours and merge after

AustinXII:
:+1:

PR #103, opened 2017-01-23
Fix Knock Off Calculation

AustinXII:


js/damage.js
```diff
@@ -131,6 +131,12 @@ function getDamageResult(attacker, defender, move, field) {
     var typeEffect2 = defender.type2 ? getMoveEffectiveness(move, defender.type2, attacker.ability === "Scrappy" || field.isForesight, field.isGravity) : 1;
     var typeEffectiveness = typeEffect1 * typeEffect2;
     var PriorityDamageCheck = move.hasPriority && (defAbility === "Queenly Majesty" || defAbility === "Dazzling") && (["Mold Breaker", "Teravolt", "Turboblaze"].indexOf(attacker.ability) !== -1) || (move.givesHealth && attacker.ability === "triage");
+    var resistedKnockOffDamage = (defender.item === "" ||
+            (defender.name === "Giratina-Origin" && defender.item === "Griseous Orb") ||
+            (defender.name.indexOf("Arceus") !== -1 && defender.item.indexOf("Plate") !== -1) ||
+            (defender.name.indexOf("Genesect") !== -1 && defender.item.indexOf("Drive") !== -1) || 
+            (defender.ability.indexOf("RKS System") !== -1 && defender.item.indexOf("Memory") !== -1) || 
+            (defender.item.indexOf("Z") !== -1));
```
MagiaSN: "Zap Plate" also has character "Z" in it so I think " Z" would be better

AustinXII: @MagiaSN hm? that doesn't make sense, also dang I was hoping I wasn't going to need to add them all individually to the variable :(

MagiaSN: @AustinXII Maybe I didn't make myself clear :/ I mean we can check `indexOf(" Z")`, in which there's a space in front of "Z", so that we can match all Z-Crystals and escape Zap Plate

AustinXII: Oh I didn't notice the space on mobile my bad, yeah that should work

AustinXII:
@Zarel you can merge this after you review

PR #108, opened 2017-01-27
Implement Leech Seed option 

AustinXII:


calc_bc.html
```diff
@@ -584,6 +584,14 @@
                             <input class="btn-input" type="checkbox" id="lightScreenR" />
                             <label class="btn btn-xwide btn-right" for="lightScreenR">Light Screen</label>
                         </div>
+                        <div class="left" title="Has this Pok&eacute;mon been effected by Leech Seed?">
```
Zarel: "affected"

AustinXII: oops

Marty-D:
A nice touch could be if setting Liquid Ooze on the Leech Seeded side would subtract that Pokemon's maxHP/8 from calcs against the opposite side. Something to consider!

AustinXII:
@Marty-D will do, there's probably a few additions I could do that involves Liquid Ooze

PR #109, opened 2017-01-30
Add defenderTyping variable

AustinXII:


Zarel:
...I don't think this is how programming works...

AustinXII:
uhm, yeah I don't know what I was trying to do

AustinXII:
An actual way to do this would to create defender.types in the Pokémon function. It'd just check type1 and type 2

PR #110, opened 2017-01-31
Implement Pokemon.typing

AustinXII:
also fix multiscale and abilities calc

### DIFF
--- a/js/ap_calc.js
+++ b/js/ap_calc.js
@@ -343,6 +343,7 @@ function Pokemon(pokeInfo) {
         var pokemon = pokedex[this.name];
         this.type1 = pokemon.t1;
         this.type2 = (pokemon.t2 && typeof pokemon.t2 !== "undefined") ? pokemon.t2 : "";
+        this.typing = this.type1 || this.type2;
         this.rawStats = [];
         this.boosts = [];
         this.stats = [];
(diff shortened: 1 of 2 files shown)


AustinXII:
Oh I guess it would have been better to just fix that last PR

js/ap_calc.js
```diff
@@ -343,6 +343,7 @@ function Pokemon(pokeInfo) {
         var pokemon = pokedex[this.name];
         this.type1 = pokemon.t1;
         this.type2 = (pokemon.t2 && typeof pokemon.t2 !== "undefined") ? pokemon.t2 : "";
+        this.typing = this.type1 || this.type2;
```
Zarel: This is still not how programming works...

MagiaSN:
According to my understanding, because both `type1` and `type2` are string (not `null` or `undefined`) and `type1` is never empty, `type1 || type2` would simply returns `type1`, so is seems this may not work

Zarel:
^

Yes, that's what I meant when I said "This is still not how programming works..."

Honko:
I think what you're aiming for is a hasTyping function, which sounds like a good idea to me.

AustinXII:
@Honko yeah and I'll get to this eventually

PR #111, opened 2017-02-04
Fix Sunsteel Strike and Moongeist Beam

ABrambleNinja:
Moongeist and Sunsteel ignore the opponent's ability when calculating damage

### DIFF
--- a/js/damage.js
+++ b/js/damage.js
@@ -69,9 +69,12 @@ function getDamageResult(attacker, defender, move, field) {
     }
     
     var defAbility = defender.ability;
-    if ((["Mold Breaker", "Teravolt", "Turboblaze"].indexOf(attacker.ability) !== -1) || move.name === "Moongeist Beam" || move.name === "Sunsteel Strike") {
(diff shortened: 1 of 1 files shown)


js/damage.js
```diff
@@ -69,7 +69,7 @@ function getDamageResult(attacker, defender, move, field) {
     }
     
     var defAbility = defender.ability;
-    if (["Mold Breaker", "Teravolt", "Turboblaze"].indexOf(attacker.ability) !== -1) {
+    if ((["Mold Breaker", "Teravolt", "Turboblaze"].indexOf(attacker.ability) !== -1) || move.name === "Moongeist Beam" || move.name === "Sunsteel Strike") {
```
MagiaSN: The body of the if statement does not only ignore the defender's ability, but also record attacker's ability in the result message, so I think it would be better to put Moongeist Beam check in another if statement

AustinXII: ^ This

AustinXII:
Fixed and merged

PR #114, opened 2017-02-04
Fix priority check

MagiaSN:
The previous priority check
`
PriorityDamageCheck = move.hasPriority && (defAbility === "Queenly Majesty" || defAbility === "Dazzling") && (["Mold Breaker", "Teravolt", "Turboblaze"].indexOf(attacker.ability) !== -1) || (move.givesHealth && attacker.ability === "triage")
`
 has many errors:

1. It should check if the attacker's ability IS NOT Mold Breaker, instead of checking attacker's ability IS Molder Breaker; in fact this is unnecessary, because if attacker's ability is Mold Breaker, the defender's ability is already set to empty there;
2. It checks Triage last with `||` operator, means if Triage is in effect then previous condition is ignored

So I rewrite the priority check, also add check for Gale Wings and Psychic Terrain (tested). The Gale Wings check has the same "full HP issue" with Multiscale, everyone can improve it if you find a proper solution.
For Psychic Terrain I added `isGroundedForCalc` function, hope it can simply terrain check in the future

### DIFF
--- a/js/damage.js
+++ b/js/damage.js
@@ -129,10 +129,15 @@ function getDamageResult(attacker, defender, move, field) {
         }
     }
 
+    if ((attacker.ability === "Gale Wings" && move.type === "Flying" && (gen >= 7 ? attacker.curHP === attacker.maxHP : true)) ||
+            (attacker.ability === "Triage" && move.givesHealth)) {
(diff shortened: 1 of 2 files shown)


AustinXII:
:+1

PR #115, opened 2017-02-04
Add Endeavor

MagiaSN:
Endeavor is added by accident when we add z-power, but its basic data (such as power) is not added yet, it seems this causes honkalculate to crash so I added it
Do we need to implement its effect? I tried to implement it but the KO chance message is really ridiculous. Or should we remove it?


AustinXII:
The ZP is probably useful, I don't see any reason to remove the move unless it's causing issues. I agree that the KO-chance will probably be ridiculous but there's no rush to get it done anyways.

AustinXII:
For now this fixes a bug so merging

PR #118, opened 2017-02-07
Add usage-based sets in place of absent analyses.

Honko:
Add usage-based sets for Pokemon in tiers where they have decent usage but no Smogon analysis. This is mainly for SM
where most analyses are still in progress, but it also fills in some gaps in old-gen lower tiers that lack full sets of
analyses and have usage stats available from the RoA Rotational ladder.


AustinXII:
@Honko some of the Pokémon are getting weird set names, such as ``RU DON&#39;T USE AMBIPOM`` It would probably be a good idea to have your script check for special characters

PR #119, opened 2017-02-08
Implement Leech Seed recovery

MagiaSN:
Also add Liquid Ooze
Heal (or damage if attacker has Liquid Ooze) defender if attacker is seeded


AustinXII:
Can I see a screenshot of what this looks like?

MagiaSN:
@AustinXII 
Before the changes, in one-vs-one just the top margin of Leech Seed button disappears:
screenshot2
But in all-vs-one it's much worse, the left Leech Seed button is at the wrong position and the right one has a wrong text:
screenshot1
After the changes all buttons look normally just like Foresight button

PR #122, opened 2017-02-23
Update CAP sets

sparktrain:
- Update sets for CAP Pokemon to better reflect viable / common sets in the Gen 7 CAP metagame
- Fix some data errors
- Add sets for Tomohawk and Kerfluffle


AustinXII:
I believe this interferes with @Honko s auto set grabbing script

Honko:
Nah, updates for CAPmons are fine. My script doesn't try to update their sets, so this block at the end never gets touched.

AustinXII:
Thanks 👍

PR #124, opened 2017-02-27
Implement protect check

MagiaSN:
Add "Protect" option, if checked, moves without `bypassesProtect` property with do no damage, and z-moves will get their damage reduced
Implement `isZ` as an own property of z-moves instead of adding it dynamically, so even if the user type the z-move themselves, it still works
Move Feint to gen 4 when it was introduced, and change its power in gen 5

### DIFF
--- a/calc_bc.html
+++ b/calc_bc.html
@@ -594,6 +594,16 @@
                             <label class="btn btn-xwide btn-right" for="lightScreenR">Light Screen</label>
                         </div>
                     </div>
+                    <div class="btn-group gen-specific g2 g3 g4 g5 g6 g7">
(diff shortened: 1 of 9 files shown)


AustinXII:
looks good 👍

AustinXII:
@MagiaSN if you're interested feel free to join the dev room on PS where you'll be given something for your contributions :)

MagiaSN:
@AustinXII Thanks, I think I'm there :)

AustinXII:
@MagiaSN if you see Zarel or me ( I just go by Austin) please pm one of us, I've seem to have just missed you

PR #132, opened 2017-03-13
Fix setting level based on the format

asgdf:
- properly set level to 50 if VGC
- immediately set proper level upon selecting set, to prevent useless
  "level changed" popups when switching sets


AustinXII:
:+1

PR #134, opened 2017-03-14
Add Gen 7 Soul Dew

ih8ih8sn0w:


AustinXII:
I could have sworn I already fixed this... anyways considering you're an official researcher, do the mega's get the soul dew boost if they're holding soul dew from mods? If not remove that check imo

ih8ih8sn0w:
Yeah, I made sure that megas got the boost as well before adding them. It was actually the main reason why I did this.

PR #140, opened 2017-03-31
Add missing contact flags and Z move powers

asgdf:
Also turned a handful of double quotes into single ones because that's what all the other strings are using.


AustinXII:
Thanks!

PR #142, opened 2017-04-08
Bulldoze and Earthquake halve damage, not base power, in Grassy Terrain

urkerab:
In particular, the base power never gets affected by Technician. https://www.smogon.com/forums/threads/3586701

Note Bulbapedia contradicts itself: some pages say half power, others say half damage, which may have contributed to the confusion.


AustinXII:
Thanks

AustinXII:
I believe you broke offline development with this commit

urkerab:
Oops, I only meant to commit the one file; I didn't realise I had garbage lying around. Sorry about that.

PR #148, opened 2017-06-16
give rse medicham sets pure power

sulcata:
probably not a good idea to edit manually, but who knows where these magical sets come from

p.s. i think pure power is a really useful ability :(


AustinXII:
If you want a set changed, change the Pokémons set analysis on smogon

PR #150, opened 2017-06-26
Improve accessibility for screen readers #149

### LINKED ISSUE #149: Improve accessibility for screen readers
Similar to [the corresponding issue for the Showdown client](https://github.com/Zarel/Pokemon-Showdown-Client/issues/954). Unfortunately, this will be much harder than the client, as that only needed usability improvements.

* [ ] The generation radio buttons are not recognized as such, they just read as plain text. There is no indication which is selected, or that they are a logical group. Might be because <input> elements need to always be inside a `<form>`, or associated with one using an attribute, so this looks like invalid html5. Similarly for the notation and mode radio buttons on the top, the move selection, single/doubles, terrain selection, weather selection, and spike layers radios.
* [x] What's this used for? `<span id="damageValues">(211, 214, 216, 218, 220, 224, 226, 227, 231, 234, 235, 238, 242, 243, 246, 250)</span>`
* [ ] When the page loads, below pokemon1, I hear "Abomasnow (ORAS NU Mixed Attacker)" being announced as a link. Only by pressing enter do I hear my focus moving into an edit field. I can edit the text, but it doesn't read suggestions when pressing up/down, or announce the presence of suggestions. When I press escape to stop editing, I hear 2 edit boxes now, one of them is probably not meant for editing, so should be `aria-hidden`.
* [x] The crit and z checkboxes are not recognized as such, but read as plain text, with no indication of being checked/unchecked. Similarly, the gravity, stealth rocks, reflect, light screen, protect, leech seed, Foresight, Helping Hand, Friend Guard and Aurora Veil checkboxes. 
* [ ] For controls that can affect either side of the field (e.g, hazards), should be labeled so it differentiates between yours and opponent's side.

Neurrone:
Fixes #149.

* [x] Change btn-input style so that it is hidden visually, but not from screen readers. This allows screen readers to see the standard HTML5 `<input>` elements being used for checkboxes and radios.
* [ ] Make the purpose of this more obvious for screen readers. (I assume there's some sort of tooltip for sighted users to explain what this does, or it is clear from visual context). `<span id="damageValues">(211, 214, 216, 218, 220, 224, 226, 227, 231, 234, 235, 238, 242, 243, 246, 250)</span>`
* [ ] Apply `aria-hidden` to an edit box not meant for users to use.
* [ ] Fix autocomplete suggestions not being read for select2 edit boxes.
* [ ] For controls that can affect either side of the field (e.g, hazards), should be labeled so it differentiates between yours and opponent's side.
* [ ] Make the calculation result an aria live region, so that any change causes that update to be immediately announced.
* [ ] Better label move selection radio buttons so that it is clearer which pokemon the move is for, since this is only obvious visually.


Zarel:
The `(211, 214, 216, 218, 220, 224, 226, 227, 231, 234, 235, 238, 242, 243, 246, 250)` is displayed with no hint to sighted users about what it means. I assume some sighted users ignore it, and others are aware that it's a list of possible damage values.

MagiaSN:
For the 2nd task, perhaps just change it to `Possible damages: (211, 214, 216, 218, 220, 224, 226, 227, 231, 234, 235, 238, 242, 243, 246, 250)` is enough?

Neurrone:
@MagiaSN perfect.

Zarel:
I'd go with "Possible damage amounts" to be clearer.

AustinXII:
I agree with possible damage amounts

Zarel:
I'm really not a fan of hiding stuff offscreen. Maybe hide it under the element or something.

Zarel:
(Hiding stuff offscreen interacts poorly with various mobile browsers and rendering engines.)

Neurrone:
What about this css [From a recent article](http://www.coolfields.co.uk/2016/05/text-for-screen-readers-only-updated/)

```css
.screen-reader-text { 
   clip: rect(1px, 1px, 1px, 1px); 
   height: 1px; 
   width: 1px; 
   overflow: hidden; 
   position: absolute !important;
}
```

AustinXII:
That looks good

Zarel:
For future reference, PS prefers rebase commits to merge commits.

Zarel:
Thanks!

Neurrone:
@Zarel oh, wasn't done with this PR yet. Will continue in part 2.

AustinXII:
Yeah was going to say I didn't think this was done

PR #151, opened 2017-06-26
Add missing some bullet flags

asgdf:
For Zap Cannon, Rock Blast, Searing Shot and Beak Blast.

Is there a reasom the calc doesn't just grab the data files from PS? Or if offline usage is important, couldn't they be auto-generated from the PS files with a script every now and then?


AustinXII:
No feel free to work on that if you want

PR #152, opened 2017-06-30
Improve accessibility for screen readers part 2

### LINKED ISSUE #149: Improve accessibility for screen readers

Neurrone:
Fixes #149

### DIFF
--- a/ap_calc.css
+++ b/ap_calc.css
@@ -365,7 +365,7 @@ select.toxic-counter {
     font-size: 0.8em;
 }
 
-.hide {
+[hidden], .hide {
     display: none;
 }
 
--- a/ap_calc.css
+++ b/ap_calc.css
@@ -243,23 +243,23 @@ th {
 .btn-mid {
     border-radius: 0;
 }
-.btn-input:focus + .btn, /* Add visual indication when focusing on visually hidden elements */
+.visually-hidden:focus + .btn, /* Add visual indication when focusing on visually hidden elements */
 .btn:hover {
     z-index: 10;
     border-color: #777777;
     background: #e3e3e3;
 }
-.btn-input:checked + .btn {
+.visually-hidden:checked + .btn {
     font-weight: bold;
     color: #111111;
     background: #f5f5f5;
 }
-.btn-input[type="radio"]:checked + .btn {
+.visually-hidden[type="radio"]:checked + .btn {
     cursor: default;
 }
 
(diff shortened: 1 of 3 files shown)


Zarel:
I mean, I don't see a huge need to do this all at once? If you prefer it that way, though, I'll leave this open.

Neurrone:
@Zarel thanks, yeah prefer to do it all at once.

Neurrone:
@Zarel ready for review/merging. Could you check particularly that no unintentional UI changes were introduced?

Zarel:
If @AustinXII would like to check it. I personally don't work on the damage calculator and am not familiar with its codebase, so I'm not a good person to review it.

Zarel:
(I merged your other one because the changes were sufficiently superficial that I could understand them eaily)

AustinXII:
I'm not a huge fan of the field selection button layout after this change

Neurrone:
@AustinXII would it suffice to somehow have the column titles hidden?

AustinXII:
@Neurrone that would probably be preferred

Neurrone:
@AustinXII done.

AustinXII:
Looks good, thanks :+1:

PR #153, opened 2017-07-03
Rename calc_bc.html to honkalculate.html

AustinXII:


Zarel:
Um, did you really just do these renames without updating the HTML files?

PR #158, opened 2017-07-28
Fix ko chance of Pokemon with leftovers and black sludge after being knocked off 

AustinXII:


AustinXII:
Ugh there looks like there's some white space, I'll fix that when I'm back on a computer either tonight or tomorrow

js/ko_chance.js
```diff
@@ -71,10 +71,11 @@
             eotText.push('hail damage');
         }
     }
-    if (defender.item === 'Leftovers') {
+    var noItem = defender.ability !== "Sticky Hold" && attacker.move === "Knock Off";
```
AustinXII: Ugh also I didn't push my commit that reversed these two >:(

Zarel:
Feel free to merge when you're ready.

Marty-D:
Uh, this doesn't do anything because `attacker.move` is not a thing. You'll notice nothing else in that file tries accessing the attacker's moves because it's just not supported until you add support.

AustinXII:
Oh whoops thought I pushed my last commit

AustinXII:
@Marty-D sorry for the wait, hows this

Marty-D:
Well, I mean, this just makes all the attacker's moves stop calculating with Leftovers if Knock Off is one of them. I'm not sure that's any better than having to manually remove Leftovers from the opponent when you want to calc with Knock Off.

AustinXII:
@MartyD Yeah I'm aware it was doing that, I was going under the assumption "If your Pokemon has knock off you probably would have used it to get rid of the opppnents item." It's probably possible to narrow it down to just the attackers selected move although I will have to look into adding that because I don't think it exist atm.

AustinXII:
after a discussion with Marty we decided this needs too much work and effort to be added correctly

PR #163, opened 2017-08-15
Update setdex_sm.js

Yoshizilla315:
Choice Band Dugtrio, while not the best set, is still relevant enough to be included, since extra damage on Earthquake could be necessary on some teams. Also, it allows you to run Aerial Ace, which is to hit Tapu Bulu and Mega Heracross, as well as Serperior. Pursuit allows you to revenge kill Latios after a Draco Meteor, which does kill 100% of the time after rocks. Like previously mentioned, not the best set, but is still a niche set for teams weak to the aforementioned Pokemon.


AustinXII:
Sorry if you want a set added you need to add it to the Pokemon smogon Analysis where honko rips these sets from

PR #165, opened 2017-08-17
Fix Searching for Mega Pokemon

AustinXII:


AustinXII:
@Honko when this is merged you will have to update your script.

AustinXII:
@Zarel @Marty-D any objections?

Zarel:
you know, you can just update the search algo instead of changing every resource file ever...

AustinXII:
Yes I'm aware but I was having issues with making that work. Although this took a while I think it's more convenient and makes searching a lot easier. I like having all the mega Pokemon show up when typing "mega" before Meganium

Honko:
Move search already does something really similar. You can find "Hidden Power Fire" or "Mystical Fire" by just typing "Fire". Would the same thing not work here? I think it would be convenient to be able to search for megas with either option (start typing "Chariz..." OR start typing "Mega C..."). I know some people are complaining about the way it is now, but there will also probably be people complaining if you change it.

I'm also not a big fan of the part where I have to update my script. Please let me be lazy. :(

AustinXII:
@Honko you will still be able to find the mega Pokemon by typing the base forme first also I think this looks athestically better as well

Zarel:
I told you, update the search algorithm.

http://dex.pokemonshowdown.com/

The Dex does that.

AustinXII:
@Zarel Oh I thought you were making a suggestion, can you link me to we're the dex does this? The dex's logic looks a lot more complex.

AustinXII:
And also makes meganium show first >:/ but I suppose I will have to compromise

Zarel:
Oh, it wasn't exactly a suggestion; I'm very opposed to changing every resource file ever, away from PS's standards, which will make future imports and general data sharing a lot more problematic.

Just manually change it.

The Dex's method is rather complicated and strewn across multiple files, easier for me just to describe it:

- autogenerate aliases based on various rules
- search an index that contains aliases as well as the original names

PR #167, opened 2017-08-20
Fix searching for Megas

Honko:
@AustinXII I took a look and I think this will do what you wanted for https://github.com/Zarel/honko-damagecalc/pull/165


AustinXII:
Aye thanks now I don't have to rush my weekend :), btw I was actually looking at this exact line earlier seeing what it meant by "remove second check when Mega's are merged"

PR #168, opened 2017-08-26
Add the pledge moves

AustinXII:
**DO NOT MERGE**
**TODO** Fix the BP mod check


AustinXII:
Hm does combining the pledge moves add another STAB?

Marty-D:
Yes; https://dex.pokemonshowdown.com/moves/firepledge

Re: this implementation, why not just take a page out of Flash Fire's book and make moves like "Fire Pledge (combo)" and forget about using even more checkboxes? Then you can handle perma-STAB and no Gem use more easily.

AustinXII:
I'm going to close this for now and push Marty suggestions when I have time

PR #174, opened 2017-09-27
Ring Target calculation fix

Ntonio36:


AustinXII:
The code styling needs fixing, also I would have done this differently, why not just check for the defenders item on any of the type effectiveness damage calcs?

Ntonio36:
It would be kind of the same story, since Ring Target denies immunities on moves while moves are single-typed, then that item will only negate one immunity out of 2 immunities theorically possible (since there is no way a Pokémon would be dual-typed Ghost/Ghost for example). Plus, it might not be known if the move type is effective on the other type (let's say Superpower on a Ring Target Sableye, it still is super effective since fighting is x2 on darkness, which shouldn't be ignored), hence the not-so-classy way of `switch`ing through typeEffectivenesses 1 or 2 to check who from them equals 0. I still can make it a traditional if / else if though, just wanted to point my point of view (woah) out.

AustinXII:
Why not just ``if (typeEffectiveness === 0 && defender.item === ‘Ring Target’) { 
typeEffectiveness = 1
}``
?

Ntonio36:
Because as I said, you still have to take into account the move type vs the secondary type of the defender (the type that isn't affected by Ring Target).

AustinXII:
Oh I see, I was misunderstanding. Anyways your code styling needs fixing

AustinXII:
I was thinking it was the same as thousand arrows but they’re different

Ntonio36:
Making it an if else should make it for good, right ?

AustinXII:
Yes this is better but by “styling” I meant your code syntax. You’re missing spaces after ``If`` and before your curly braces

Ntonio36:
The code is implemented and works, the styling has been fixed, what's left ?

AustinXII:
Sorry, thanks for that,  I’m just going to test and make sure this works how intended. Give me a few hours

AustinXII:
Thanks!

Issue #121, opened 2017-02-10
To Do list


AustinXII:
I figured putting a todo list on the GitHub repo itself might be beneficial as it's much easier for developers to see what still needs to be done.
``1: Fix Mega's/alolan Pokémon search`` currently the calculator doesn't support searching for the Pokémons Mega or Alola forms. If you wanted to search for mega metagross for example, you'd have to type "metagross-mega". Changing the ``getSetOptions`` function in ``ap_calc.js`` should be the solution here 
``2: Fix best move auto selection`` currently when you change a pokemons EVs after already doing a calc it will redirect you off of the move that was previously the strongest. This is annoying people so adding a lock-move option might be a good thing to have, the best being if you double click on a move it stays highlighted.
``3: Add forme-change option`` a dropdown option that changes the Pokémons forme but not the set. This would require adding ``hasAltForm`` in ``pokedex_js``
``4: Imported sets only mode`` an option where only your imported sets show up on the left side of the calc, while the other showing sets from ``setdex_**.js``
``5: Fix stat dropping moves KO-Chance`` auto calculate ko chance with moves that drop a users stats. Pretty straight forward, add  ``dropsStats`` to any relevant move in ``moves.js`` then change any relevant code in ``ko_chance.js``

AustinXII:
@MagiaSN feel free to tackle any of these if you're interested :)

MagiaSN:
@AustinXII I will, just have something else to do recently :)

AustinXII:
1 and 3 done in #157

Ntonio36:
Can I join in ? 6 looks easy so I probably will be able do deal with it.

AustinXII:
This is open to anyone :) feel free

AustinXII:
2 was done as well

Ntonio36:
Can you strike through the list elements that have been made, please ?

AustinXII:
Yeah @Ntonio36 last time I tried the syntax didn’t work for somebody reason

Ntonio36:
N°6 done in #180 and #181

Ntonio36:
Number 8 done in #190

Tyler-Hilbert:
Can we add in an export tool for all the presets?  Like it would be nice if I could just pull a pokemon off of damage calculator and put it into my team.

Ntonio36:
Export option added in #217.

PR #177, opened 2017-10-07
Add Parabolic Charge

Ntonio36:
Was checking on Bulbapedia for moves that drain HP (for #121 issue number 6) , and didn't find this one when looking it up in the moves search in the calculator.


AustinXII:
This is wrong, if you’re going to use bulbapedia please also look at when a move was first introduced and what changes happened to the move, IE BP

Ntonio36:
Man it's okay I'm sorry xD

PR #181, opened 2017-10-16
Fix visual disorder

Ntonio36:
Draining HP at x4 effectiveness will make the main description text go wild to the right while the percentage of the related move goes for an additional line, ruining the view. This fix reduces `recoveryText`'s length so even if the percentage went really high, nothing would be messed up.

To understand what the issue is about, try Giga Draining with Abomasnow on a Swampert for example.

EDIT : and no, no formatting issue.

### DIFF
--- a/js/index_controls.js
+++ b/js/index_controls.js
@@ -46,7 +46,7 @@ function calculate() {
                 10 : Math.floor(minDamage * p1.moves[i].percentHealed * 48 / p1.maxHP);
             var maxHealthRecovered = notation === '%' ? Math.floor(maxDamage * p1.moves[i].percentHealed * 1000 / p1.maxHP) / 
                 10 : Math.floor(maxDamage * p1.moves[i].percentHealed * 48 / p1.maxHP);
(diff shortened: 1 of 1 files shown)


Ntonio36:
Sorry, found out everything was ok. My bad. ^^""

Ntonio36:
Or not, after all.

AustinXII:
Sorry I can’t test rn can you just screenshot what you’re fixing?

Ntonio36:
[Before](https://screenshots.firefox.com/I31gxkQcqE1Wq863/null)
[After](https://screenshots.firefox.com/QbxwEyfa1hNzmD8P/null)

AustinXII:
Thanks!

Zarel:
@Ntonio36 

> EDIT : for anyone seeing this past October 2017, the images are unavailable.

Have you considered just attaching the pictures to GitHub so it's not a problem?

Zarel:
**Before**

image

**After**

image

Zarel:
Like this (also I cropped it so it could be easier to see which part you fixed)

Ntonio36:
@Zarel thank you !

PR #186, opened 2017-11-19
Fix Photon Geyser/Light That Burns The Sky damage calculation

asgdf:


AustinXII:
This isn’t correct, “Photon Geyser compares the user's Attack and Special Attack stats, dealing damage to the target using the higher stat.”

AustinXII:
Oh well this is right but I mean can you also try to implement the desc

asgdf:
Alright, now it does the "use Atk or SpA depending on which is higher" thing too. I was initially under the impression that that was already implemented.

AustinXII:
Yeah I forgot to do this, thanks!

PR #187, opened 2017-11-22
Add Travis

AustinXII:


AustinXII:
This was mostly copied over from PS, so please check over this and add/fix whatever, 
crash emails will also have to be changed.

AustinXII:
@Zarel

Zarel:
You have to run `npm test` and fix all the errors before Travis CI will be useful to you.

package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "pokemon-showdown-damage-calc",
+  "private": true,
+  "description": "The official damage calculator for the Pokémon Showdown battle simulator",
+  "version": "0.3.0",
+  "scripts": {
+    "start": "node pokemon-showdown",
+    "test": "eslint --rulesdir=dev-tools/eslint --cache"
```
Zarel: This should be `eslint --cache .`

Zarel:
```
✖ 145373 problems (145324 errors, 49 warnings)
  141509 errors, 2 warnings potentially fixable with the `--fix` option.
```
At least they're nearly all fixable? Although it probably means that you need to set up `.eslintignore` correctly.

AustinXII:
Oh when the Travis test didn’t fail so I thought it was fine, will fix

Zarel:
There won't be a Travis test until this pull request is merged. There definitely isn't a "passed Travis" message anywhere on this pull request.

Zarel:
In hindsight, I should probably be the one doing this...

AustinXII:
I meant on the Travis website for my build when I merged these commits on my github... but yeah you probably should be as I’ve never worked with Travis before @Zarel

AustinXII:
I did at least try to test this before merging, I probably should have done a console ``npm test`` though

Zarel:
So a lot the question here is stuff like, what do we want our standard to be? Right now, some files are tabs, some files are two spaces, some files are four spaces.

Also: switchcase indentation: Do we want to match Main? (Main doesn't indent `case` clauses; the damagecalc currently does)

Also: comma spacing: It's currently inconsistent; mostly, commas are spaced in most places except arrays. Should we keep this (eslint won't be able to check it if so) or just make them all spaced?

Don't worry about the amount of work of these changes; `eslint` can change them automatically.

Zarel:
Also: I understand now. The Travis build said "no errors" because you didn't set `npm test` to actually check the code, you set it to just open the eslint help, which it managed to do without any errors.

AustinXII:
@Zarel how do you use eslint to change them automatically?

Zarel:
`eslint --fix ...` but you might as well just tell me and I'll do it myself.

AustinXII:
@Zarel Ah well the code styling ultimately is your decision so I’m really fine with whatever. I think tabs would be better for the file style, switch case indentation should probably match main and we should just make all commas are spaced.

Zarel:
Okay, doing that.

Zarel:
I fixed all the style errors except for `no-undef`, and pushed.

https://github.com/Zarel/honko-damagecalc/commit/87992ed67805a9fbb1c4e8e553b09abf5432108a?w=1

Here's the non-whitespace changes, which you might want to review. You probably also want to restructure the indent pyramids eslint seems to prefer into switch statements or something: https://github.com/Zarel/honko-damagecalc/commit/87992ed67805a9fbb1c4e8e553b09abf5432108a?w=1#diff-178a90500caccbce7b63e50e6bed3037R533

Zarel:
Now all you need is to fix the 578 global accesses that you can see with `npm test`. A lot of those are actual bugs (forgetting to use `var`), and the remaining ones, well, the short way is to list all your globals in `globals`, but what I usually actually do is use namespaces (e.g. `Data.SETDEX_RBY` instead of `SETDEX_RBY`). You should generally have relatively few globals.

Zarel:
You still haven't actually fixed the global accesses...

js/damage_dpp.js
```diff
@@ -1,4 +1,5 @@
 ﻿function CALCULATE_ALL_MOVES_DPP(p1, p2, field) {
+	var SP;
```
Zarel: This is _not_ how you localize a global variable...........

Zarel: Is this just a global string constant? Just actually list it as global in eslintrc

AustinXII: Yeah I noticed that yesterday when I rebased. The reason it isn’t already was because you didn’t want too many globals, but yeah this declaration is just wrong. The other stats will also have to be fixed

Zarel: This doesn't _stop_ it from being a global.

Honestly, I don't mind having the six stat constants as globals. It's probably easier than whatever other plan you have for them.

AustinXII: Yeah that’s what’s i noticed lol i definetly got confused somewhere down the road as originally they were listed as globals but you requested them to be changed at the time but anyways will change

Zarel: I said you shouldn't make _everything_ a global. These seem to make more sense as globals.

js/moveset_import.js
```diff
@@ -24,6 +24,7 @@ function ExportPokemon(pokeInfo) {
 			EV_counter += pokemon.HPEVs;
 			EVs_Array.push(pokemon.HPEVs + " HP");
 		}
+		var stat;
 		for (stat in pokemon.evs) {
```
Zarel: These should be `for (var stat in ...`

js/moveset_import.js
```diff
@@ -36,6 +37,7 @@ function ExportPokemon(pokeInfo) {
 		finalText += "\n";
 	}
 	var movesArray = [];
+	var i;
 	for (i = 0; i < 4; i++) {
```
Zarel: `for (var i = 0; ...`

js/shared_controls.js
```diff
@@ -1,35 +1,28 @@
...
-	}
-
-	k = Math.max(n >= 0 ? n : len - Math.abs(n), 0);
-	while (k < len) {
-		if (k in O && O[k] === searchElement) {
-	    	return k;
-	  	}
-	  	k++;
-	}
+	this.indexOf = function (searchElement, fromIndex) {
```
Zarel: This should still be `Array.prototype.indexOf =`.

You can tell eslint to ignore this if it's still an issue.

js/shared_controls.js
```diff
@@ -686,10 +679,10 @@ function Side(format, terrain, weather, isGravity, isSR, spikes, isReflect, isLi
 }
 
 var gen, genWasChanged, notation, pokedex, setdex, typeChart, moves, abilities, items, STATS, calcHP, calcStat;
-
+/* eslint-disable no-global-assign */
```
Zarel: Ummm after looking it up, this is definitely not a setting that ever needs to be suppressed.

What's the problem here?

AustinXII:
@Zarel I know I finally spent some time on this again, we’re finally almost there!! Will fix your issues and start reversing a lot of the global declarations tonight

Zarel:
I actually think we might want to merge this now, and put those fixes in a separate commit.

Zarel:
For "reversing a lot of the global declarations" I'd probably do those in a separate pull request entirely.

js/damage.js
```diff
@@ -239,9 +239,10 @@ function getDamageResult(attacker, defender, move, field) {
 	}
 
 	if (move.name === "Spectral Thief") {
+		var stat;
 		for (stat in defender.boosts) {
```
Zarel: `for (var`

js/damage.js
```diff
@@ -780,9 +781,10 @@ function getDamageResult(attacker, defender, move, field) {
 		} else {
 			boostTurns = move.metronomeCount;
 		}
+		var metronome;
 		for (metronome = 0; metronome < boostTurns; metronome++) {
```
Zarel: `for (var`

js/moveset_import.js
```diff
@@ -46,16 +46,17 @@ function ExportPokemon(pokeInfo) {
...
 
-$("#exportR").click(function() {
+$("#exportR").click(function () {
 	ExportPokemon($("#p2"));
 });
 
 function serialize(array, separator) {
 	var text = "";
+	var i;
 	for (i = 0; i < array.length; i++) {
```
Zarel: `for (var`

js/moveset_import.js
```diff
@@ -107,18 +108,17 @@ function getStats(currentPoke, rows, offset) {
 		var evs = {};
 		var ivs = {};
 		var ev;
+		var j;
```
Zarel: delete this

js/moveset_import.js
```diff
@@ -107,18 +108,17 @@ function getStats(currentPoke, rows, offset) {
...
 
 		switch (currentRow[0]) {
 		case 'Level':
 			currentPoke.level = parseInt(currentRow[1].trim());
 			break;
 		case 'EVs':
-
 			for (j = 1; j < currentRow.length; j++) {
```
Zarel: `for (var`

js/moveset_import.js
```diff
@@ -248,6 +251,8 @@ function addSets(pokes) {
 	var currentRow;
 	var currentPoke;
 	var addedpokes = 0;
+	var i;
+	var j;
 	for (i = 0; i < rows.length; i++) {
```
Zarel: `for (var`

js/moveset_import.js
```diff
@@ -248,6 +251,8 @@ function addSets(pokes) {
...
 		currentRow = rows[i].split(/[\(\)@]/);
 		for (j = 0; j < currentRow.length; j++) {
```
Zarel: `for (var`

js/shared_controls.js
```diff
@@ -149,10 +142,10 @@ function drawHealthBar(poke, max, current) {
 	healthbar.addClass("hp-" + fillColor);
 	var unwantedColors = ["green", "yellow", "red"];
 	unwantedColors.splice(unwantedColors.indexOf(fillColor), 1);
+	var i;
 	for (i = 0; i < unwantedColors.length; i++) {
```
Zarel: `for (var`

js/moveset_import.js
```diff
@@ -332,20 +334,19 @@ $(bothPokemon("#importedSets")).click(function () {
 	var pokeID = $(this).parent().parent().prop("id");
 	var showCustomSets = $(this).prop("checked");
 	if (showCustomSets) {
-		loadCustomList(pokeID);
+		loadCustomList(pokeID); // eslint-disable-line no-undef
```
Zarel: ?

js/moveset_import.js
```diff
@@ -222,6 +222,8 @@ function addToDex(poke) {
 }
 
 function updateDex(customsets) {
+	var pokemon;
+	var moveset;
 	for (pokemon in customsets) {
```
Zarel: `for (var`

js/moveset_import.js
```diff
@@ -222,6 +222,8 @@ function addToDex(poke) {
...
 		for (moveset in customsets[pokemon]) {
```
Zarel: `for (var`

Zarel:
👍👍👍 This is great!

js/moveset_import.js
AustinXII: travis was complaining about loadcustomlist being undefined

PR #190, opened 2017-12-05
TODO List #8

Ntonio36:
Provides a button to clear custom sets once not needed anymore.

### DIFF
--- a/index.template.html
+++ b/index.template.html
@@ -112,6 +112,7 @@
         <fieldset class="poke-info" id="p1">
             <legend align="center">Pok&eacute;mon 1</legend>
             <input type="text" class="set-selector calc-trigger" />
+			<button id="clearSets" style="display:none">Clear Custom Sets</button>
             <div class="info-group">
                 <div>
                     <label>Type</label>
--- a/js/moveset_import.js
+++ b/js/moveset_import.js
@@ -260,9 +260,16 @@ function checkExeptions(poke) {
 
 }
 
+$("#clearSets").click(function(){
+	localStorage.removeItem("customsets");
+	alert("Custom Sets successfully cleared.");
+	$("#clearSets").css("display","none");
+});
+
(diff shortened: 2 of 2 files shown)


AustinXII:
Rad, screenshot of this?

Ntonio36:
before clearance
Regardless of how likely will you face that kind of Zekrom, all what we care about here is the "Clear Custom Sets" button that shows up instantly as the custom set is added. 

Now after clicking the button, POOF all the sets disappeared, the magic button did its job so it quietly hides back (hence the `display` set to `none` when done, we don't want from it to bloat the HTML). When looking up the custom set of Zekrom we had crafted : 
after
It isn't there anymore.
And by the way, thank you very much for the support, I just try my hardest to be what I'm here for : helping :D.

Ntonio36:
before clearance
Regardless of how legitimate this set is, this is how the page should look like once a custom set has been imported. Notice the "Clear Sets" button that is the pivotal thing in the whole story.

Once clicked, I try again to look for the custom set I've crafted for Zekrom (second image below), aaaand nope, no "Custom Set" appears, it has been deleted successfully, as the alert mentioned when the button was clicked.

Oh, and thanks for the nice word, glad I could help :D.

AustinXII:
This only clears the sets after a refresh. It would be optimal for a refresh to be unneeded but for now this is good

AustinXII:
@Ntonio36 Hmm in the request for this, they wanted specific sets, is that something you could work on as well?

Ntonio36:
I successfully found a way to separately store custom sets all alone, but I couldn't figure out at all how to make them get loaded into the Pokémon name area... That's a TODO I should set for myself, for the least that I can say. But don't expect anything from it, though.

AustinXII:
@Ntonio36 try looking at how customsets is currently done there

Issue #130, opened 2017-03-09
Speed stat ups not taken into account when calculating first pokemon to move


Gindranis:
Setting Speed to +2 does not alter the generated outcome.
Tested with various pokemon that carry a move that would deal 100%+ damage, but get outsped before +2 speed, such as cloyster vs tapu koko, Haxorus vs Garchomp, Gengar vs Alakazam

MagiaSN:
I don't quite understand this issue, the speed stats do not affect the damage directly, so speed changes are not shown in the result message

If the power of the move depends on speed stats (e.g. Electro Ball), the changes is shown directly in move BP

If you mean "when setting Cloyster's speed to +2, then Icicle Spear should be automatically chosen instead of opposite Tapu Koko's Thunderbolt, because Cloyster should move first", it seems there is no such feature now, but I think the maintainers (@AustinXII)  would be glad to take it into consideration

AustinXII:
Planned

Ntonio36:
5 months later, I'm working on it. Probably done in like 20 days ? 

Regardless, expect big things to be changed.

Ntonio36:
PR done.

AustinXII:
fixed in #192

PR #192, opened 2017-12-16
Show faster Pokémon's strongest move (fix issue #130)

Ntonio36:
It took me less than 20 days, I must recognize. But well, let me demonstrate what will happen : 

### At loadup
The left Abomasnow's Blizzard will show up. Nothing unusual.

### Lowering left Abomasnow's Speed EVs by 4
The right Abomasnow's Blizzard will show up, because it is recognized as the Pokémon who attacks first.

### Changing right Abomasnow's Giga Drain's type to Fire
Its Giga Drain will show up, since it is the strongest move (x4 effectiveness, 100% - 100% draining, etc.)

### Left Abomasnow has +6 speed boost
Its Blizzard is shown back.

### Left Abomasnow has -1 speed boost
The view is switched back into the right Abomasnow's modified Giga Drain

### Left Abomasnow becomes Deoxys-Attack (Uber All-out attacker)
Deoxys' Superpower is automatically shown since : 
- Deoxys-Attack is faster than Abomasnow
- Superpower deals the most damage on Abomasnow

(side note : probably should cap shown damage amount as well ?)

### DIFF
--- a/js/index_controls.js
+++ b/js/index_controls.js
@@ -26,8 +26,12 @@ var damageResults;
 function calculate() {
 	var p1 = new Pokemon($("#p1"));
 	var p2 = new Pokemon($("#p2"));
+	var battling = [p1, p2];
+	p1.maxDamages = [];
+	p2.maxDamages = [];
 	var field = new Field();
 	damageResults = calculateAllMoves(p1, p2, field);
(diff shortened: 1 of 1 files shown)


AustinXII:
Looks good I’ll test and merge today, just curious what’s your ps account and timezone?

Ntonio36:
My Pokémon Showdown account : MK360
Timezone : UTC+02:00 (Beirut)

AustinXII:
Thanks!

Issue #182, opened 2017-10-18
RBY/GSC mix up stat order when determining HP IV


SadisticMystic:
Nowadays, the canonical ordering of stats is HP-AT-DF-SA-SD-SP, but it was originally ordered as HP-AT-DF-SP-SC, and that's the order that's reflected in the G1 and G2 data structure. If you go into one of those modes and set the Special IV to 14, HP is also supposed to drop to 14 because the low bit there is also the low bit for HP, but the calculator actually sets HP to 13 there, as though it thinks Special controls the 2 bit of HP. That bit is really supposed to be intermeshed with Speed, and setting a 14 there should drop HP to 13 in the absence of anything else, but instead the calc goes to 14 HP--a straight relabeling of the stat order.

Ntonio36:
So, as far as I understood, the Speed DVs actually weigh 2 bits while the Special DVs only weigh one ?

AustinXII:
No I think it’s just an issue with the code not being tuned correctly for older gens since it’s mixing up the stat order. It’s presumably using gen3+ stat order in gens 1&2

Ntonio36:
@SadisticMystic I made a PR that might fix the problem(s), could you test and notify me whether this has solved the issue globally ?

Ntonio36:
@AustinXII you can close this issue now :).

AustinXII:
Fixed in #194

PR #194, opened 2017-12-18
Switch bit weights between Special and Speed stats

Ntonio36:
Closes issue #182.. I guess.

In short : instead of considering 2 bits for Special and one for Speed, the Speed DVs affect 2 units of HP DVs, while the Special ones do not affect but one.

### DIFF
--- a/js/shared_controls.js
+++ b/js/shared_controls.js
@@ -84,9 +84,9 @@ $(".sl .dvs").keyup(function () {
 
 function getHPDVs(poke) {
 	return (~~poke.find(".at .dvs").val() % 2) * 8 +
-	(~~poke.find(".df .dvs").val() % 2) * 4 +
-	(~~poke.find(".sp .dvs").val() % 2) * 2 +
-	(~~poke.find(gen === 1 ? ".sl .dvs" : ".sa .dvs").val() % 2);
+(~~poke.find(".df .dvs").val() % 2) * 4 +
+(~~poke.find(".sp .dvs").val() % 2) * 2 +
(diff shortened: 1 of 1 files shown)


js/shared_controls.js
```diff
@@ -85,8 +85,8 @@ $(".sl .dvs").keyup(function () {
 function getHPDVs(poke) {
 	return (~~poke.find(".at .dvs").val() % 2) * 8 +
             (~~poke.find(".df .dvs").val() % 2) * 4 +
-            (~~poke.find(gen === 1 ? ".sl .dvs" : ".sa .dvs").val() % 2) * 2 +
-            (~~poke.find(".sp .dvs").val() % 2);
+				(~~poke.find(".sp .dvs").val() % 2) * 2 +
```
AustinXII: Remove white space

Ntonio36: Maybe good enough if I put them all in one line, so it doesn't end up like the first recovery text PR (messy indenting everywhere) ?

AustinXII:
Hmm guess it’s not so easy on mobile, if you can’t fix it, I’ll fix it for you when I test it. You should be able to highlight white space with your mouse highlighter in case you weren’t aware.

Ntonio36:
Looks _way_ better now.

Ntonio36:
So, what now ?

AustinXII:
Gotta fix the spacing for this and I’m going to give sadisticmystic a bit more time to verify his issue is fixed

SadisticMystic:
That works.

AustinXII:
ugh annoying, but fixed lol. Also thanks for the confirmation @SadisticMystic

PR #195, opened 2017-12-19
Update dropdown Nature menus with help (e.g.: "Adamant (+Atk/-SpA)")

penpexgit:
I think this could be useful since sometimes for Natures that are not often seen one does not remember well what it does.
And for newbies it's definitely a useful resource, instead of having to check another website in the meantime.
Most other calcs have it, and it saves time at no cost, so here it is.

### DIFF
--- a/honkalculate.template.html
+++ b/honkalculate.template.html
@@ -337,31 +337,31 @@
                         <div class="gen-specific g3 g4 g5 g6 g7">
                             <label>Nature</label>
                             <select class="nature">
-                                <option value="Adamant">Adamant</option>
+                                <option value="Adamant">Adamant (+Atk, -SpA)</option>
                                 <option value="Bashful">Bashful</option>
-                                <option value="Bold">Bold</option>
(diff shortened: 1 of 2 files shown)


penpexgit:
For some reason unknown to me, it seems to now auto-select Mild, instead of Hardy, even though there is still `<option value="Hardy" selected="selected">Hardy</option>`
If anyone has an idea why, I can make the change.

penpexgit:
That's right, and it is also the default behavior currently, not sure why I just didn't realize that.
So yeah everything works fine and when you click a Blank Set, Hardy is still the default as expected.

AustinXII:
Can your scree shot this? I remember I did this at one point but I think the UI looked terrible so I abandoned it

penpexgit:
I think it looks fine, you can see for yourself [here](https://i.imgur.com/aEHnQSc.png) :)

AustinXII:
Thanks!

PR #196, opened 2017-12-20
Fix Nature Drop-down

LegoFigure11:
As per issue identified here: http://www.smogon.com/forums/threads/3593546/page-3#post-7620768


penpexgit:
Ah yeah, that was great QA from me...I guess I just checked the left part.
Thanks for the fix, now if you correct the amount of white space to match the one all other Natures have that'd be perfect.

PR #197, opened 2017-12-21
TODO List task No.4

Ntonio36:
At last it has arrived!

## Demo screenshots
Given this custom set :
```
Reshiram @ Draco Plate
Modest Nature
- Fusion Flare
- Energy Ball
- Thunderbolt
- Ice Beam
```
This is what happens to the page upon successful importing (notice the new checkbox) : 

imported sets

Till now, nothing fancy. But when you check the box and you wanna select a Pokémon from there.... (the white Kyurem was already there)

dropdown

Our Reshiram appears.

Now the reason why a blank select value shows up first is explained in both `shared_controls`' commits.

But then, the most interesting part, when you finally click a custom set : 

reshiram

_Tadaaa_

Now as a side note, I'm not quite sure if anyone agrees with the select default value being blanked for default sets. But I just hope everything goes well. 🤷‍♂️

### DIFF
--- a/index.template.html
+++ b/index.template.html
@@ -112,7 +112,10 @@
         <fieldset class="poke-info" id="p1">
             <legend align="center">Pok&eacute;mon 1</legend>
             <input type="text" class="set-selector calc-trigger" />
-			<button id="clearSets" style="display:none">Clear Custom Sets</button>
+            <span id="importedSetsOptions" style="width:auto; display:none">
+                <input type="checkbox" id="importedSets" /> Only show imported sets <br />
+                <button id="clearSets">Clear Custom Sets</button>
+            </span>
             <div class="info-group">
                 <div>
                     <label>Type</label>
--- a/js/moveset_import.js
+++ b/js/moveset_import.js
@@ -300,8 +300,7 @@ $("#importedSets").click(function () {
 	var showCustomSets = $(this).prop("checked");
 	if (showCustomSets) {
 		loadCustomList();
-	}
-	else {
+	} else {
 		loadDefaultList();
 	}
 });
@@ -314,8 +313,7 @@ $(document).ready(function () {
 		customSets = JSON.parse(localStorage.customsets);
 		updateDex(customSets);		
 		$("#importedSetsOptions").css("display","inline");
-	}
-	else {
+	} else {
 		loadDefaultList();
 	}
 });
--- a/js/moveset_import.js
+++ b/js/moveset_import.js
@@ -289,16 +289,33 @@ function checkExeptions(poke) {
 
 }
 
-$("#clearSets").click(function(){
+$("#clearSets").click(function () {
 	localStorage.removeItem("customsets");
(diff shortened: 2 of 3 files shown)


AustinXII:
Yay!! Thank you :)

AustinXII:
Awesome no spacing errors either, just watch the ``else`` formatting :+1:

penpexgit:
Maybe it's just me, but I tried with Firefox, Chrome, IE, with both the online version of the calc, and the local one by loading the index.html, but the checkbox never appears when I import the set given in example (or any other set of course).
Everything else works as described though.
Is it just me, or something is wrong with he most important feature of this commit?

AustinXII:
Ugh I didn’t test that when I was making sure this worked, @Ntonio36 pls fix

Ntonio36:
Tried it back out, seems like the changes didn't sync to the online page, since I get the expected results on a local copy..

Microsoft Edge doesn't support passing default values to function arguments (`shared_controls.js`, line 740), but seriously, why using Edge?
As for the local copy not showing the applied changes, are you sure you redownloaded the most recent version? I have zero problem with Firefox: everything behaves like it's supposed to do. Currently downloading Chrome, I'll test it back in a few moments.

(side unrelated note : GitHub still allows me to bring my branch back alive, much wow)

AustinXII:
Yeah github has some cool features, thanks for investigating although we should have support for all browsers. If possible can you refactor?

Ntonio36:
Yes sir, needs some time but I should be able to handle it.

AustinXII:
Rad thank you

penpexgit:
Ok so I don't know if you tested on Chrome or not @Ntonio36, or if @AustinXII tested this and got the same result as me (?) but here are some (weird) updates:

1) I retried the live version: still the same as yesterday, just "Clear custom sets" shows up, not the checkbox on Firefox, Chrome, IE.
2) I had git cloned the repo to test locally yesterday so I just thought since you said that it works on your side, hey I'll dowload the zipped master branch directly from Github. Well the results are different, but still not what they should be: the index.html loaded in Firefox or Chrome doesn't even show Abomasnow on my side, the line is just blank, but it is Abomasnow that is selected, since the calcs show that. When I import a set, nothing happens, BUT when I reload the page, then the checkbox shows up, and then the behavior is as expected...!
3) I tried again git cloning and loading up the index.html to be sure but got the exact same behavior as yesterday, which was no checkbox on any browser.

So I'm not sure what's going on exactly...if more tests could be done by others on many browsers, it would be great. Maybe reopen the PR for visibility?

Zarel:
For the record, I'm intentionally leaving the live version unupdated until this thing gets fixed; please tell me when it does.

Ntonio36:
@penpexgit found the source of the problem for your point 1 and fixed it. 
Point 2 : this is the expected behavior. Define "this" as "the line being blank but the calculator still showing Abomasnow".
Now it's all done and repaired, check #198.

PR #198, opened 2017-12-23
TODO List, Task No.4 patch

Ntonio36:


Zarel:
You should really learn to rebase so the commit list isn't so ugly.

AustinXII:
If you need any help learning how to rebase feel free to message me or anyone in the PS development room ``.com/dev``

Ntonio36:
@AustinXIII ready and set to merge (thanks @Zarel !)

js/damage.js
```diff
@@ -493,14 +493,13 @@ function getDamageResult(attacker, defender, move, field) {
 		description.attackerAbility = attacker.ability;
 	}
 
-	if ((attacker.item === "Thick Club" && (attacker.name === "Cubone" || attacker.name === "Marowak" || attacker.name === "Marowak-Alola") && move.category === "Physical") ||
+	if ((attacker.item === "Thick Club" && (["Cubone", "Marowak", "Marowak-Alola"].indexOf(attacker.name) !== -1) ||
```
AustinXII: You still need to check for physical moves here

AustinXII:
I would prefer the ``damage.js`` refactoring be done in a different PR if you can do that

js/damage.js
```diff
@@ -360,18 +362,14 @@ function getDamageResult(attacker, defender, move, field) {
...
 	} else if (attacker.item === move.type + " Gem") {
 		bpMods.push(gen >= 6 ? 0x14CD : 0x1800);
 		description.attackerItem = attacker.item;
-	} else if (((attacker.item === "Soul Dew" && attacker.name === "Latios") ||
-        (attacker.item === "Soul Dew" && attacker.name === "Latios-Mega") ||
-        (attacker.item === "Soul Dew" && attacker.name === "Latias") ||
-        (attacker.item === "Soul Dew" && attacker.name === "Latias-Mega")) &&
-        (move.type === attacker.type1 || move.type === attacker.type2)) {
+	} else if (attacker.item === "Soul Dew" && ["Latios", "Latias", "Latios-Mega", "Latias-Mega"].indexOf(attacker.name) !== -1 &&
+		isSTAB) {
```
AustinXII: I’d just put this with the line above even though it’s getting pretty long

Ntonio36:
To be honest, now that it's "done", I'd prefer not to mess with it again. Also, I suppose creating a separate PR would create a conflict in `damage.js`, and I guess no one likes to face problems, especially when in the day before Christmas :D.

AustinXII:
That is true but now the work that you’ve done has actually caused issues that you will have to spend time on fixing which is what I was afraid of :>

AustinXII:
Now this will have to be put on hold for another day :/

Ntonio36:
I do have time, just point me at where the problems are.

AustinXII:
Viewing the changes section on this PR should make things pretty obvious, there’s quite a bit

Ntonio36:
Was wondering what was that dark red background for, my goodness. Good thing it's over now.

js/shared_controls.js
```diff
@@ -825,52 +829,52 @@ function getTerrainEffects() {
...
-	case "type2":
-	case "item":
-		var id = $(this).closest(".poke-info").prop("id");
-		var terrainValue = $("input:checkbox[name='terrain']:checked").val();
-		if (terrainValue === "Electric") {
-			$("#" + id).find("[value='Asleep']").prop("disabled", isGrounded($("#" + id)));
-		} else if (terrainValue === "Misty") {
-			$("#" + id).find(".status").prop("disabled", isGrounded($("#" + id)));
-		}
+		case "type1":
```
AustinXII: Can you fix these errors?

Zarel:
Honestly, this is why you need to set up Travis CI, @AustinXII :p

AustinXII:
Soon, holidays are hectic :o

js/shared_controls.js
```diff
@@ -835,7 +839,7 @@ function getTerrainEffects() {
 		} else if (terrainValue === "Misty") {
 			$("#" + id).find(".status").prop("disabled", isGrounded($("#" + id)));
 		}
-		break;
+	break;
```
Zarel: Your `break` indentation is still wrong.

AustinXII:
Thanks!

penpexgit:
So I just tested the last version of this, and since it is clearly not live yet on the online calc, I tried on the `index.html` of the cloned version, and everything seems to work as expected, but I have one remark:

@Ntonio36 , you said [here ](https://github.com/Zarel/honko-damagecalc/pull/197#issuecomment-353714855) that, at startup, having a blank line for the Pokémon on each side is the intended behavior (instead of showing the two Abomasnow)?
I find it odd, since the stats, moves, and calcs (so basically everything else) shows that it is indeed two Abomasnow that are taken into account.
I think that for newbies, it makes the calc less intuitive, and that it should show the two Abomasnow as before.
Maybe it's just me but I'd like to hear other people's opinion before defacing the homepage of the calc.

AustinXII:
I don’t find it bad, you can still clearly tell which 2 Pokémon are being calc’d and considering 99.9% of the time you’re going to immediately change the Abomasnow v Abomasmow matchup, I like this addition

Zarel:
I agree with penpexgit here, it should be clearer.

Zarel:
Also I suggest starting with Unown, because of the general pun factor.

penpexgit:
Just to add some arguments and make my point more explicit:
1) I think that the fact of having blank lines instead of showing the two Abomasnow removes information from the user, and _might_ make the page less readable to a total newbie
2) The main thing is that it brings nothing to have a blank line. You _still_ have to click one time to change your Pokémon so I really don't see how it _adds any value_ with respect to the previous behavior to have blank lines...

So one drawback, and no gain in my opinion.

I concur to the idea of having Unown as default Pokémon, that's a good pun :)

penpexgit:
Maybe this should be reopened, since it seems that all the commits following the ones linked to this PR have not been pushed to the online calc (I assume it is linked to what @Zarel said in #197 [here](https://github.com/Zarel/honko-damagecalc/pull/197#issuecomment-353696415)), the calc is still stuck in the bastardized bugged state where the checkbox does not even show up.
Once the blank line thingy discussed above is fixed, I assume the changes can be pushed so we have an updated online calc.

AustinXII:
I fixed this issue @penpexgit

penpexgit:
Ah, I guess even though I follow all commits, I didn't see a modification concerning this pass, thanks for fixing it. I should have cloned the new version anyway to make sure, the blank line has now been removed indeed.
Out of curiosity, is it this [commit](https://github.com/Zarel/honko-damagecalc/commit/9f671013d00adce8a0ba8afae886488a59d2540b) in `moveset_import.js` that fixed it? (Since it was bundled with all this other stuff, I didn't see it :-) )

Also: what about the clever idea of putting Unown as a starter Pokémon? :)

AustinXII:
nah @penpexgit  I just fixed it myself and the Unown idea may happen at a later date for now this works.

penpexgit:
Sounds good, but is the fix invisible on the list of commits then?

Also, again out of curiosity, how is it chosen to push the changes of this repo to the online version, and how often does it happen? Is it just manual, when enough changes have been made?

Zarel:
I assume this is the commit? 7553aa4f517189

Can you test to see if it works?

penpexgit:
I had tested it after this commit was made, and as said above, it indeed works!
I just didn't think this corresponded to it :)

AustinXII:
Yeah I tested it

AustinXII:
So you can update whenever :P

PR #205, opened 2018-01-02
Remove Soul Dew SpA boost for Z-moves

Articuno-I:


Zarel:
Why does this matter?

Articuno-I:
It doesn't affect any calculations. But the code looks neater, makes more sense, and uses the same number of tests. Sorry for not making this clear in the initial pull request.

AustinXII:
There are two soul dew checks iirc. Your code styling is also meh

Zarel:
Does the `isZ` check need to exist at all?

AustinXII:
@Zarel Not unless an OM comes out that allows Z-moves to be used without a Z-crystal and allows Lati@s with soul dew.

AustinXII:
Usually ``lati@s + soul dew`` is put in the banlist though, at least for XY

AustinXII:
And even at that point the Z move calculation would be entirely wrong as it gives a BP of 1 to any Z-Move without a move to inherit.

AustinXII:
@Zarel oh you mean in general, yes

Zarel:
Specifically, removing `!move.isZ &&` from this code does not seem like it would cause any problems.

AustinXII:
@Zarel yeah this definitely doesn’t matter at all. It would only be relevant if an OM came out that allowed Lati@s with soul dew and z moves without crystals.

AustinXII:
Heh I’m having some déjà vu as I once made a PR to allow soul dew on mega lati@s in the beginning XY era. That was originally denied but later added down the road, so with that in mind maybe we should allow this.

Zarel:
I actually recommend removing `!move.isZ &&`, because at best it does nothing, at worst it'll interfere with an OM.

AustinXII:
Yeah I can agree with that, so this can be closed this

js/damage.js
```diff
@@ -501,8 +501,9 @@ function getDamageResult(attacker, defender, move, field) {
             (attacker.item === "Light Ball" && attacker.name === "Pikachu") && !move.isZ) {
 		atMods.push(0x2000);
 		description.attackerItem = attacker.item;
-	} else if ((gen < 7 && attacker.item === "Soul Dew" && (attacker.name === "Latios" || attacker.name === "Latias") && move.category === "Special") ||
-            (!move.isZ && (attacker.item === "Choice Band" && move.category === "Physical" || attacker.item === "Choice Specs" && move.category === "Special"))) {
```
AustinXII: Oh @Zarel you wanted this one removed? There’s probably others then can be removed as well but there are some that are actual mechanics so I’ll probably go through this unless Articuno wants to

Zarel: Yeah, I did.

AustinXII: Hmm I think I recall a complaint about choice specs boosting z-moves so actually this check needs to stay

AustinXII: @Zarel Yeah I was right https://github.com/Zarel/honko-damagecalc/commit/3cdb78644fbafb005ca2472f93d93bac67af8e04 these are needed

Zarel: They shouldn't be. You should maybe instead display a warning that the combination is not naturally allowed.

AustinXII: There were complaints that Z-moves don’t get boosted Ingame from items. I don’t think we should cater to non existent OM’s over actual in game mechanics especially when people are going to complain why it was changed. Also, there isn’t really a good place to put a warning message unless you’re usung alerts, which most people probably don’t want. TL;DR we should only complicate things for our selfs if needed.

Zarel: We aren't, though? Like if you input a hypothetical that's not possible in-game, you should either simulate that hypothetical correctly or display an error. Displaying an incorrect result is entirely the wrong solution.

Zarel: The only reason Z-moves don't get boosted from items is because item + Z-crystal is not possible in-game. So you should either be displaying a "not possible" error, or simulating the situation as if it were possible. Instead, you're displaying an incorrect number. The worst of both worlds! And you're making the code even more complicated than necessary, to do it!

AustinXII: The calc wouldn’t be wrong until the Z-moves get boosted though, which it doesn’t. It’s hypothetically giving you the correct damage if you modded your cartridge to allow for these cases. I only added the checks because people pointed out the calculations were wrong.

Zarel: Or am I misunderstanding something? Is this how it works in Hackmons?

AustinXII: Unless someone gave me the wrong info when complaining, this is following in game mechanics, which disallows Z-moves from item boosts even if it hypothetically was possible.

AustinXII: We’ll need a researcher because I can’t find any info on the web.

Zarel: @Marty-D

Zarel: PS does not seem to implement this, and this would be pretty easy to implement, so I am doubting the truth of this.

AustinXII: Here’s one of the request to clarify and it got liked by a decent amount of people http://www.smogon.com/forums/threads/pokémon-showdown-damage-calculator-request-thread.3593779/page-2#post-7279192

Zarel: That's asking for convenience, not making a normative statement about game mechanics.

Zarel: And the recommended implementation is to negate the held item, not add a bunch of `if` statements. :|

AustinXII: There were more somewhere that was just an example. But yeah it probably could have been better I just did it that way while people were still pointing out mechanics. Anyways I’m fine with removing it if these aren’t actual mechanics

Zarel: I do see merit in that request. The solution is to negate the held item.

AustinXII: He specified life orb and like items and I was unsure how z-moves interacted with other items at the time. Regardless if it was more simplified that line would have to be removed as we’re finding out it’s wrong

AustinXII: We can’t have both? Are you changing what you prefer done?

Zarel: Yeah, I'm changing it because people asked for something else in the linked post. Do you just not understand what they're asking for?

AustinXII: No I thought you still wanted what you asked originally. I fully understand what they want I just did it they way I did it because new mechanics were still being discovered daily and I was unsure how z-moves would interact with all items at the time, as stated before.

Zarel: No, they're not requesting anything relevant to mechanics, they're requesting an ability to quickly switch between their chosen item and a Z-crystal, for convenience purposes.

AustinXII: Again there were other requests, that was just an example. I may of confused mechanics with a request originally though, it’s hard to remember everything from a year ago :p.

AustinXII: But anyways there was a reason for it, it can just be optimized now.

Marty-D: Of course item modifiers boost Z-Move damage, it's just not possible normally.

PR #206, opened 2018-01-03
Additional fix

Ntonio36:
Deals with damage calculation inaccuracy and defender-helping stuff (Leftovers, gained Leech Seed health) text not showing up.

I know everyone is getting fed up, this was my whole fault, I have to concede.

### DIFF
--- a/js/damage.js
+++ b/js/damage.js
@@ -570,7 +570,7 @@ function getDamageResult(attacker, defender, move, field) {
 	////////////////////////////////
 	//////////// DAMAGE ////////////
 	////////////////////////////////
-	var baseDamage = Math.floor(Math.floor((Math.floor((2 * attacker.level) / 5 + 2) * basePower * attack) / defense) / 50 + 2);
+	var baseDamage = getBaseDamage(attacker.level, basePower, attack, defense);
 	if (field.format !== "Singles" && move.isSpread) {
 		baseDamage = pokeRound(baseDamage * 0xC00 / 0x1000);
 	}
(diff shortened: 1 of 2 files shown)


AustinXII:
Gonna to have to call in the big guns for this, @Marty-D pls review :>

Ntonio36:
Could anyone add a _do not merge_ tag please ? There still are some issues I'm working on.

Ntonio36:
Aaaaand done, now awaiting reviews and approvals.

AustinXII:
@Ntonio36 can you resolve the merge conflicts?

Ntonio36:
@penpexgit happy now?

penpexgit:
Well, the numbers are good now for the calcs so that's good!

As for the result text, I see you opted for my number 2 option that I detailed [here](https://github.com/Zarel/honko-damagecalc/pull/200#issuecomment-354706828), where option 3 had my preference (btw, I realize now that my option 1 is identical to my option 3 in the end).
So to make it clear to everyone, since I think some discussion on this topic would be nice instead of a unilateral decision (I didn't see any except my aforementioned post), here is what happens now (this is "option 2"):

Starting with -1 SpA, and executing 2 times Draco Meteor will show:
`-3 252+ SpA Life Orb Abomasnow Draco Meteor over 2 turns vs. 0 HP / 0 SpD Abomasnow: 247-293 (76.9 - 91.2%) -- not a KO`
So it shows the overall stat modifier at the beginning of the second move. This is better that what was in place before this fix for sure (it would have shown -5 in this case, mixing a stat modifier _before_ the move was executed, and a stat modifier _after_ the move was executed, we can call it "option 0" if somebody wants to discuss it), but I feel like "option 2" is still not the optimal choice here.

Instead, my preferred choice, "option 1/3", would show this:
`-1 252+ SpA Life Orb Abomasnow Draco Meteor over 2 turns vs. 0 HP / 0 SpD Abomasnow: 247-293 (76.9 - 91.2%) -- not a KO` and it would not change for any other number of times the move is used.
It would just show (if any) the stat modifier at the beginning of the sequence of any multi-turn stat-dropping move and consider this sequence as one move, a single entity: (2/3/4/5 times Draco Meteor). This would definitely remove any possible confusion as to whether the stat shown is the one _after_ or _before_ the stat drop, and also make it consistent with how _all_ the other moves are treated.

The purpose of the result text after all is to give (in the minimum space possible) ALL the data taken into account unambiguously to make the calc, and give the result in terms of HP and % chance of KO. _Unambiguously_ being the term relevant to our discussion here, which rules out "option 2". Also, the result text does not need to address the state of the attacker after an attack, which is why it also rules out "option 0" (which brings also even more ambiguity of course, cf what was said above).

Ntonio36:
Hmm I guess you're right, after all. Retouched on that point and it's about to be committed in a while. Thanks for the feedback :).

PR #208, opened 2018-01-04
Refactor Recoil Text

AustinXII:


Marty-D:
>Since `hasRecoil` exists already, you can change them from `true` to `50` or `33` or `25` and just use that instead of all this.

I meant, like, actually repurpose it, i.e.
```javascript
'Double-Edge': {
	bp: 100,
	type: 'Normal',
	category: 'Physical',
	makesContact: true,
	hasRecoil: 33
},
```
and
```javascript
if (p1.moves[i].hasRecoil) {
	var minRecoilDamage = notation === '%' ? Math.floor(Math.min(minDamage, p2.maxHP) * p1.moves[i].hasRecoil * 10 / p1.maxHP) / 10 :
		Math.floor(Math.min(minDamage, p2.maxHP) * p1.moves[i].hasRecoil * 0.48 / p1.maxHP);
	var maxRecoilDamage = notation === '%' ? Math.floor(Math.min(maxDamage, p2.maxHP) * p1.moves[i].hasRecoil * 10 / p1.maxHP) / 10 :
		Math.floor(Math.min(maxDamage, p2.maxHP) * p1.moves[i].hasRecoil * 0.48 / p1.maxHP);
	recoilText = ' (' + minRecoilDamage + ' - ' + maxRecoilDamage + notation + ' recoil damage)';
}
```

Ntonio36:
Or, instead of `hasRecoil`, maybe rename it to `recoilPercent`?

AustinXII:
whoops @Marty-D fixed. Also added checks for highjumpkick and the likes.

js/index_controls.js
```diff
@@ -116,35 +107,29 @@ function calculate() {
...
+			var maxRecoilDamage = notation === '%' ? Math.floor(Math.min(maxDamage, p1.maxHP) * p2.moves[i].hasRecoil * 10 / p2.maxHP) / 10 :
+				Math.floor(Math.min(maxDamage, p1.maxHP) * p2.moves[i].hasRecoil * 0.48 / p2.maxHP);
+			if (minDamage > p1.curHP) {
+				minRecoilDamage = notation === '%' ? Math.floor(Math.min(p1.maxHP * p2.moves[i].hasRecoil) * 10 / p2.maxHP) / 10
+				: Math.floor(p1.maxHP * p2.moves[i].recoilPercentage * 48 / p1.maxHP);
+				maxRecoilDamage = notation === '%' ? Math.floor(Math.min(p1.maxHP, p2.moves[i].hasRecoil) * 10 / p2.maxHP) / 10
+				: Math.floor(Math.min(p1.maxHP, p2.moves[i].hasRecoil) * 0.48 / p2.maxHP);
+			}
 			recoilText = ' (' + minRecoilDamage + ' - ' + maxRecoilDamage + notation + ' recoil damage)';
+		} else if (p2.moves[i].hasRecoil && typeof p2.moves[i].hasRecoil !== 'number') {
```
Marty-D: No need to check that it's not a number here, since getting to this point at all means it's not a number. Also keep in mind if you decide to show crash damage text, it's different in almost every gen.

js/data/move_data.js
```diff
@@ -94,7 +94,7 @@ var MOVES_RBY = {
 		type: 'Normal',
 		category: 'Physical',
 		makesContact: true,
-		hasRecoil: true
+		hasRecoil: 33
```
Marty-D: I guess using Double-Edge as my example was a bad idea since its recoil is 25% before Gen 3.

js/index_controls.js
AustinXII: I did it like this so every calc doesn’t get recoil text, but checking if hasRecoil check is probably enough

js/data/move_data.js
AustinXII: Whoops oh yeah lol

js/index_controls.js
AustinXII: (To clarify, checking if it wasn’t a number wasn’t enough as every calc got recoil text at that point)

AustinXII:
Alright fixed

js/index_controls.js
Marty-D: Yeah, I got that, I was only talking about the number check.

You seem to have also added a check for Mind Blown with `hasRecoil` without accounting for it in Reckless calculations (it's not boosted).

AustinXII: What other changes are there? I looked at the dex on PS and double edge was the only one?? And oh I forgot about reckless, this is why I didn’t want to mess with hasRecoil :> Also I was talking about doing the else statement that way, but yeah you were right

AustinXII: Are you talking about struggle? Struggle isn’t on the calc currently but I am going to do it, just don’t know how much time I’ll have this week and for now this works

Marty-D: The text I'm talking about is the one associated with this block, not the one above. I also mentioned it in the first comment here. Bughouse posted the exact specifications for crash damage in each gen in your thread.

AustinXII: Oh I haven’t checked smogon since Friday and ah I finally get what you mean. Easy enough fix, I’ll do it when I do struggle. Btw did you say crash damage at first? I read recoil damage so I thought you meant the line above / double edge again, my bad.

AustinXII: (You definitely said crash damage, curse adhd I gotta learn to read more thoroughly oml)

PR #211, opened 2018-01-05
Update instructions in README

asgdf:


AustinXII:
You could also just download the zip file from github and double click index

asgdf:
Am I missing something
There is no index.html in the repository

AustinXII:
There’s a index template file

AustinXII:
Hmm I guess it should specify that for that option

asgdf:
You can open the template file I suppose, but you won't be able to open the 1vsAll/Allvs1 mode unless you actually build them

AustinXII:
Or you could use the honkaculate template file

asgdf:
jQuery throws an error if you do that

Zarel:
Let's just fix that one issue, then? I'd rather tell people to open the template file, than install a bunch of otherwise-unnecessary dependencies for a build.

asgdf:
Okay so I looked into this, the jquery error was just because there's no mode URL parameter at the end if you just open the html file manually.
Now, the problem is that the controls in the top right to switch modes reference the non-template versions. This means that if you have opened honkalculate.template.html, you can't switch between the 1vsAll/Allvs1 modes unless you manually modify the URL.

Would it be sensible to have the mode controls link to the template files if you are already using a tempate file, and to the "proper" versions otherwise?

Alternatively, the instrutions could just say "create copies of the template files and rename them", you don't need a build script for that part after all.

Zarel:
> Would it be sensible to have the mode controls link to the template files if you are already using a tempate file, and to the "proper" versions otherwise?

Sounds good to me.

asgdf:
Okay, how's this? Switching modes works with both templates and built files now. Also, opening the honkalculate.html file directly works now because I specified a default mode.

README.md
```diff
@@ -13,7 +13,7 @@ Most notable for having all-vs-one (wall) and one-vs-all (sweeper) calcs.
 Instructions
 ------------
 
-Double-click `index.html`
+Double-click `index.html` or, if that file doesn't exist, `index.template.html`.
```
Zarel: Just "Double-click `index.template.html`"

js/shared_controls.js
```diff
@@ -936,6 +936,18 @@ function loadCustomList() {
 		});
 }
 
+function isUsingTemplates() {
+	return  (window.location.href.substring(window.location.href.lastIndexOf('/')).indexOf('.template.html') >= 0);
+}
+
+function get1vs1Page() {
```
Zarel: ...this should really be done by the template compiler, not the code.

Like just update `./build` to find/replace `.template.html` with `.html`, then the rest of this code is unnecessary.

js/honkalculate_controls.js
```diff
@@ -295,7 +295,11 @@ $(".set-selector").change(function (e) {
 var mode, dtHeight, dtWidth;
 $(document).ready(function () {
 	var url = window.location.href;
-	mode = url.substring(url.indexOf('=') + 1, url.length);
+	if (url.indexOf('=') < 0) {
```
Zarel: Cache search functions if you're reading from them more than once.

js/shared_controls.js
asgdf: But wasn't the whole point to not force people to run the build script

js/honkalculate_controls.js
asgdf: Is it worth doing that for two function calls? I just took this as the easy way out to avoid having to think of a variable name ("indexOfEqualsSign"? idk)

js/shared_controls.js
Zarel: Exactly!

js/honkalculate_controls.js
Zarel: I'd call it `equalsPos`.

But then again, I've written this code millions of times. Good to get into the right habit.

js/shared_controls.js
Zarel: Specifically, you make the links to `.template.html`, so they work without the build script, and then the build script replaces `.template.html` with `.html`, so it works in the live version.

Which is what I _thought_ you were going to do until you dropped all these unnecessary functions. :p

asgdf: oohhhhh I get it now

asgdf:
Finally got around to this - so you mean like this?

I have to say I'm not really a fan of having a script modify the code like this, but I suppose it works

build
```diff
@@ -21,6 +21,12 @@ function makeCachebuster(template, output) {
 		try {
 			const filename = c.replace('//play.pokemonshowdown.com/', '../../play.pokemonshowdown.com/');
 			const fstr = fs.readFileSync(path.resolve(rootDir, filename), {encoding: 'utf8'});
+			if (fstr.indexOf('.template.html') >= 0) {
```
Zarel: You don't need to replace every `fstr`, only `indexContents`.

js/honkalculate_controls.js
```diff
@@ -294,8 +294,13 @@ $(".set-selector").change(function (e) {
 
 var mode, dtHeight, dtWidth;
 $(document).ready(function () {
-	var url = window.location.href;
-	mode = url.substring(url.indexOf('=') + 1, url.length);
+    var url = window.location.href;
```
Zarel: Indentation

Zarel:
Yes, you're nearly there!

build
Zarel: ...In hindsight, you do need to replace every `fstr`.

But you still need to also replace every `indexContents`, unless it already does the thing.

Zarel: ...which I guess they don't. Nevermind, then.

Zarel:
Oh, the problem is that these files are versioned. Blah...

Zarel:
The new way is to just define a `linkExtension = '.template.html';` directly in `index.template.html`, so then you only have to replace them in `indexContents` rather than in the JavaScript files.

asgdf:
I like that solution, mostly because it doesn't actually modify the files itself
Does this look good?

The indentation thing you mentioned was spaces surrounded by tabs, at tab width 4 it didn't look any different - there were also two other instances of that, so I fixed those as well

AustinXII:
@Zarel

Zarel:
Oh, I didn't see that you'd already done it.

PR #212, opened 2018-01-05
fixed recoil ranges for rolls above 100%

cantsay:


AustinXII:
My other PR is fine, I’ll just change the If statement to the one we agreed one and add the else if. There still is some other things this PR is missing that I was going to add in mine so I’m closing this.

AustinXII:
Sorry I got confused, you said you were going to edit my PR but I’m just realizing that’s impossible as you aren’t a collaborator. Mb for the mixup

PR #214, opened 2018-01-06
Add Nature's Madness, fix Eviolite

Ntonio36:
Didn't think adding one move would require that much editing.. Glad it's done, though!

### DIFF
--- a/js/damage.js
+++ b/js/damage.js
@@ -225,12 +225,12 @@ function getDamageResult(attacker, defender, move, field) {
 	}
 
 	if (move.name === "Guardian of Alola") {
-		var zLostHP = defender.isProtected ? Math.floor(defender.curHP / 4) : Math.floor(defender.curHP * 3 / 4);
+		var zLostHP = field.isProtected ? Math.floor(defender.curHP / 4) : Math.floor(defender.curHP * 3 / 4);
 		 return {"damage": [zLostHP], "description": buildDescription(description)};
 	}
 
 	if (move.name === "Nature's Madness") {
-		var lostHP = defender.isProtected ? 0 : Math.floor(defender.curHP / 2);
+		var lostHP = field.isProtected ? 0 : Math.floor(defender.curHP / 2);
 		return {"damage": [lostHP], "description": buildDescription(description)};
 	}
 
@@ -243,6 +243,7 @@ function getDamageResult(attacker, defender, move, field) {
 	////////// BASE POWER //////////
 	////////////////////////////////
 	var basePower;
+
 	switch (move.name) {
 	case "Payback":
 		basePower = turnOrder === "LAST" ? 100 : 50;
--- a/js/damage.js
+++ b/js/damage.js
@@ -211,7 +211,7 @@ function getDamageResult(attacker, defender, move, field) {
 
 	description.HPEVs = defender.HPEVs + " HP";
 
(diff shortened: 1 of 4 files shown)


js/damage.js
```diff
@@ -211,7 +211,7 @@ function getDamageResult(attacker, defender, move, field) {
 
 	description.HPEVs = defender.HPEVs + " HP";
 
-	if (move.name === "Seismic Toss" || move.name === "Night Shade") {
+	if (["Seismic Toss","Night Shade"].indexOf(move.name) !== -1) {
```
Zarel: Does `Array#indexOf` exist on older browsers? You may want to polyfill if you haven't.

Ntonio36: According to Mozilla Developer Network, this method is natively supported since IE9. Does anyone use an older browser that IE9, out of curiosity?

Zarel: Usually, the problem isn't IE9, it's Safari 5.

Also, why not support IE8? Polyfilling isn't a big deal; client does it.

Ntonio36:
Polyfill done.

Ntonio36:
Instead of endlessly scrolling through the whole `pokedex` file, I've added a property `canEvolve` for each Pokémon that can evolve, and the calc bases itself off of this property to decide whether Eviolite defense boost should be applied or not (because right now, you can make Reshiram get lesser damage if it holds it, which doesn't quite make sense, you'll agree with me).

js/shared_controls.js
```diff
@@ -1,3 +1,39 @@
+if (!Array.prototype.indexOf) {
```
AustinXII: @Zarel is this what you wanted?

AustinXII:
If @Zarel 's concerns are fixed, this is good to go

js/shared_controls.js
```diff
@@ -553,7 +589,10 @@ function getZMoveName(moveName, moveType, item) {
 														moveName === "Thunderbolt" && item === "Aloraichium Z" ? "Stoked Sparksurfer" :
 															moveName === "Thunderbolt" && item === "Pikashunium Z" ? "10,000,000 Volt Thunderbolt" :
 																moveName === "Volt Tackle" && item === "Pikanium Z" ? "Catastropika" :
-																	ZMOVES_TYPING[moveType];
+																	moveName === "Nature\'s Madness" && item === "Tapunium Z" ? "Guardian of Alola" :
+																		moveName === "Nature\'s Madness" && item !== "Tapunium Z" ? "Nature\'s Madness" :
+																			ZMOVES_TYPING[moveType];
+																			// What if we had Nature's Madness and the user checked the Z move option? The calc will be tricked into Twinkle Tackle.
```
Zarel: um

Ntonio36: Nvm ^^".

js/shared_controls.js
Ntonio36: Since both modes (1 vs All / All vs One / 1v1) use that method in their codes, I believed putting that polyfill at the very top of the file (which is the first file called after the datas) would be good to deny eventual problems in the future, although I'm pretty sure the browser will go through that condition wherever I would place it...

Ntonio36:
How about now?

js/shared_controls.js
Ntonio36: It is obsolete now, see the next commit.

AustinXII:
sorry for the wait, thanks!

PR #215, opened 2018-01-08
Add Struggle damage calculation 

AustinXII:


AustinXII:
the math needs to be fixed but this is mostly done.

Marty-D:
Instead of adding new things and duplicating even more recoil calculations, you can use `hasRecoil` for the appropriate Struggle values, then change it to `hasRecoil: 'struggle'` in DPP and remove the Reckless check for it, since it conveniently stopped acting like a damage-based recoil move as soon as Reckless started existing.

And similarly, `hasRecoil: 'crash'` for the Jump Kicks. Then the Reckless check is just a check for a numeric value or 'crash'.

AustinXII:
Something like this, @Marty-D ?

Marty-D:
Yep, looking good. 👍

js/data/move_data.js
```diff
@@ -355,6 +355,11 @@ var MOVES_RBY = {
 		bp: 0,
 		type: 'Grass'
 	},
+	'Struggle': {
+		bp: 1,
```
Marty-D: Should be 50. Were you thinking of zp maybe?

AustinXII: Ooh I was looking at pp lmao that makes a lot more sense, I was wondering why it was so low

js/index_controls.js
```diff
@@ -79,8 +79,28 @@ function calculate() {
 				: Math.floor(p2.maxHP * p1.moves[i].hasRecoil * 0.48 / p1.maxHP);
 			}
 			recoilText = ' (' + minRecoilDamage + ' - ' + maxRecoilDamage + notation + ' recoil damage)';
+		} else if (p1.moves[i].hasRecoil === 'crash') {
```
AustinXII: @Marty-D how’s this block for mechanics?

js/index_controls.js
```diff
@@ -79,8 +79,28 @@ function calculate() {
...
+			var minRecoilDamage = notation === '%' ? Math.floor(Math.min(minDamage, p2.maxHP) * 10 / p1.maxHP) :
+				Math.floor(Math.min(minDamage, p2.maxHP) * 0.48 / p1.maxHP);
+			var maxRecoilDamage = notation === '%' ? Math.floor(Math.min(maxDamage, p2.maxHP) * 10 / p1.maxHP) :
+				Math.floor(Math.min(maxDamage, p2.maxHP) * 0.48 / p1.maxHP);
+			if (minDamage > p2.curHP) {
```
Marty-D: Not really sure what this part is about, can you explain?

js/index_controls.js
```diff
@@ -79,8 +79,28 @@ function calculate() {
...
+				minRecoilDamage = notation === '%' ? Math.floor(p2.maxHP * 10 / p1.maxHP) / 10
+				: Math.floor(p2.maxHP * 0.48 / p1.maxHP);
+				maxRecoilDamage = notation === '%' ? Math.floor(p2.maxHP * p1.moves[i].hasRecoil * 10 / p1.maxHP) / 10
+				: Math.floor(p2.maxHP * 0.48 / p1.maxHP);
+			}
+			var gen4CrashDamage = Math.floor(p2.maxHP * 0.5 / p1.maxHP * 100);
+			recoilText = gen === 1 ? ' (1hp damage on miss)' :
+			gen === 2 ? (p2.type1 === "Ghost" || p2.type2 === "Ghost") ? ' (no crash damage on Ghost types)' : ' (' + minRecoilDamage * 12.5 + ' - ' + maxRecoilDamage * 12.5 + notation + ' crash damage)' :
```
Marty-D: Gen 2's crash damage isn't capped at the target's max HP. If you're level 100 and you miss against something that's level 2, you probably just KOed yourself.

AustinXII: Oh that’s funny, I’ll remove the cap

js/index_controls.js
Marty-D: Shouldn't all the checks for the opponent's maxHP be for curHP then?

Marty-D: The checks for normal recoil moves, I mean.

AustinXII: Yeah I thought the same thing I just didn’t get a chance to test that and make sure it works. Should be done tomorrow

Marty-D: Sounds good. Anyway, I think the way to calculate crash damage would be just to take the full min/max damage and cap it depending on the gen and/or the user's actual HP, then deal with the fancy notation bits after. Am I missing something?

AustinXII:
@Marty-D maybe something like this instead?

Marty-D:
Sure, if it works that sounds fine.

AustinXII:
I have no idea why the gen check isn't working, but everything else is fine :x

PR #218, opened 2018-01-16
Fix forme abilities

Ntonio36:
For a long time, this problem would remain unsolved.

The basic problem was that upon switching and playing around formes (most prominently between a Mega form and its "basic" one), the ability would sometimes not show up properly, and this problem would make Abomasnow (PU Showdown Usage) never get its Snow Warning despite it was listed in the abilities dropdown AND mentioned in the set ability. Now thanks to a very valuable heads-up from @AustinXII, this is now from the past.


AustinXII:
Dope I told ya this was going to be a way easier way than what you originally had planned. Good job!

Ntonio36:
I should probably learn to listen to experienced people first before trying to sort things out myself 😆 .

PR #219, opened 2018-01-17
Fix crash on non-existent item in set

### LINKED ISSUE #213: When loading a custom set on the right side, Abomasnow form remains, messing up calc
I followed the following steps:
Load this set:
```
Necrozma-Dusk-Mane @ Solganium Z  
Ability: Prism Armor  
EVs: 252 Atk / 4 Def / 252 Spe  
Adamant Nature  
- Sunsteel Strike  
- Photon Geyser  
- Prismatic Laser  
- Brick Break  
```

then load this set:
```
Gengar  
Ability: Cursed Body  
IVs: 0 Atk  
- Will-O-Wisp 
```

Both imported with success.
Then on the left side, I select 'Only show imported sets', then select my Necrozma-Dusk-Mane set.
Then on the right I select my Gengar manually from the list (since _there is no 'Only show imported sets ' option on the right_).
Then the calc shows this:
`252+ Atk Necrozma-Dusk-Mane Photon Geyser vs. 0 HP / 0 Def `**Abomasnow**`: 584-690 (181.9 - 214.9%) -- guaranteed OHKO`
Yes, Abomasnow, and in the right field, everything has Gengar's stuff, _except_ the 'Forme' field, that still has Abomasnow, so the error night come from here. After that if you try to change the Pokémon on the right to like Abra or some other, and then reselect my Gengar set, a lot of weird stuff happens, sometimes it will work, sometimes the previous Pokémon will still show up...
So I'm not sure what's happening exactly but it is not right. And yeah, **would be nice to have the 'Only show imported sets' option on the right side too.**

Ntonio36:
Closes #213 (again).
The reason why the calc would show Abomasnow as described in the issue was because the defender's set had no item, and the item is needed to check whether the Pokémon in question can mega-evolve or not (as far as I saw, at least). But because there was no item - nothing to compare, then - the calc would simply shout that `set.item is undefined`.

### DIFF
--- a/js/shared_controls.js
+++ b/js/shared_controls.js
@@ -386,7 +386,7 @@ $(".forme").change(function () {
 		setName = fullSetName.substring(fullSetName.indexOf("(") + 1, fullSetName.lastIndexOf(")"));
 
 	$(this).parent().siblings().find(".type1").val(altForme.t1);
-	$(this).parent().siblings().find(".type2").val(altForme.t2 === undefined ? "" : altForme.t2);
+	$(this).parent().siblings().find(".type2").val(altForme.t2 ? altForme.t2 : "");
 	$(this).parent().siblings().find(".weight").val(altForme.w);
 
 	for (var i = 0; i < STATS.length; i++) {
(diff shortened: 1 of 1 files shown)


Ntonio36:
Wait, I'll do a cleaner PR for it.

PR #220, opened 2018-01-17
Fix calc crash if Pokémon doesn't hold any item

### LINKED ISSUE #213: When loading a custom set on the right side, Abomasnow form remains, messing up calc

Ntonio36:
Closes #213

### DIFF
--- a/js/shared_controls.js
+++ b/js/shared_controls.js
@@ -356,16 +356,17 @@ function showFormes(formeObj, setName, pokemonName, pokemon) {
 
 	if (setName !== 'Blank Set') {
 		var set = setdex[pokemonName][setName];
-
-		// Repurpose the previous filtering code to provide the "different default" logic
-		if ((set.item.indexOf('ite') !== -1 && set.item.indexOf('ite Y') === -1) ||
-            (pokemonName === "Groudon" && set.item.indexOf("Red Orb") !== -1) ||
-            (pokemonName === "Kyogre" && set.item.indexOf("Blue Orb") !== -1) ||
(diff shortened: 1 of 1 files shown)


AustinXII:
This reminds me, I need to add mega stones to ``items.js``

PR #221, opened 2018-01-17
Implement terrain seeds

Ntonio36:


js/damage.js
```diff
@@ -854,6 +856,22 @@ function checkDownload(source, target) {
 		}
 	}
 }
+
+function checkSeedBoost(holder, field) {
+	var mappedField = field.getSide();
+	if (mappedField.terrain && holder.item.indexOf("Seed") !== -1) {
+		var terrainSeed = holder.item.substring(0, holder.item.indexOf(" "));
+		console.log(terrainSeed);
```
AustinXII: Probably don’t need this

AustinXII:
Uh? are you still testing your code? Please test everything, this literally doesn't do anything

js/damage.js
```diff
@@ -854,6 +856,21 @@ function checkDownload(source, target) {
 		}
 	}
 }
+
+function checkSeedBoost(holder, field) {
```
AustinXII: wtf is holder? attacker works?

Ntonio36:
proof
You can clearly see that it does work for me.

js/damage.js
Ntonio36: If you paid attention to the Pokémon passed as the argument for the function upon calling it, you will see `p1` and `p2`, not `attacker` or `defender`. Because you might just as well equip both Pokémon with an Electric Seed on an Electric Terrain, for purely arbitrary reasons.

AustinXII: this still doesn't answer my question, p1 is the same as attacker. wtf is holder.

AustinXII:
@Ntonio36 could you update the KO chance text to reflect this? It's easy to miss the change. That or have it just auto change the df/sd stats

js/damage.js
Ntonio36: Uh.... holder points at the Pokémon that holds the item?

AustinXII: no I mean where else is holder used, why use "holder" when attacker/p1 or defender/p2 means the same thing?

PR #222, opened 2018-01-18
Add Mega Stones

Ntonio36:
Sorry @AustinXII 🤷‍♂️ 😂


AustinXII:
there should be more than just the items being added. Knock off calc's etc.

Ntonio36:
The new commit also fixes #223 and #224, so make sure to close these issues on merge.

js/damage.js
```diff
@@ -685,7 +686,6 @@ function getDamageResult(attacker, defender, move, field) {
 	var finalMod = chainMods(finalMods);
 
 	var damage = [];
-	var cumulatedStatDrops = 0;
```
AustinXII: Did you mean for this line to be removed?

Ntonio36: Yup, it was some sort of leftover variable I had forgotten to remove when I redid the multi-turn damage calc logic.
Don't worry though, this variable is used nowhere so we can peacefully get rid of it.

PR #229, opened 2018-01-27
Updates

Ntonio36:


js/moveset_import.js
```diff
@@ -253,19 +254,17 @@ function addSets(pokes) {
 				currentPoke = POKEDEX_SM[currentRow[j].trim()];
 				currentPoke.name = currentRow[j].trim();
 				currentPoke.item = getItem(currentRow, j + 1);
-				if (j === 1) {
-					currentPoke.nameProp = currentRow[j - 1].trim();
-
+				if (j === 1 && currentRow[0].trim()) {
```
AustinXII: ???

Ntonio36: If the nickname-seeking regex finds and effectively splits the name part into this: `Nickname (Species) @ Item`, just make sure the found nickname isn't a blank name (if someone would import `(whateverspecies) @ whateveritem`, show the `whateverspecies` entry as `whateverspecies (Custom Set)` and not `whateverspecies (Nickname)`.

PR #231, opened 2018-02-15
Add Metronome

Ntonio36:


AustinXII:
@Zarel I screwed up and didn’t specify metronome the item to Antonio, what’s your opinions on this? I think this is fine but with the chances for this to generate the same move as you generate in game/ PS is incredibly unlikely I can understand why you wouldn’t want this

Zarel:
Yeah this doesn't seem useful to me.

Ntonio36:
If the problem is only about the moves and how do they get generated, erm it's okay it's not quite a big deal, maybe I should imitate the PRNG implemented in the sim?
Although yeah I feel like if we wanna calculate any move, just insert it in the Pokémon's moveset :shrug:

AustinXII:
Sorry again for the confusion

PR #233, opened 2018-02-18
Add Metronome item

Ntonio36:


js/damage.js
```diff
@@ -761,6 +761,21 @@ function getDamageResult(attacker, defender, move, field) {
...
+		var boostTurns;
+		if (move.dropsStats) {
+			boostTurns = move.usedTimes;
+		} else {
+			boostTurns = move.metronomeCount;
+		}
+		for (metronome = 0; metronome < boostTurns; metronome++) {
+			var totalMetronomeBoost = 1 + metronome / 10;
+			damage = damage.map(function(damageRoll) {
+				return pokeRound(damageRoll * totalMetronomeBoost);
```
AustinXII: this is wrong, this causes the calc to multiple the roll of metronome used once by the consecutive metronome hit. So we are currently doing ``1.0, 1.1, 1.1*1.2, 1.1*1.2*1.3, ...`` instead of ``1.0, 1.2, 1.4, 1.6, 1.8, 2.0``. I would image there is a much simpler way to do this without looping

PR #236, opened 2018-03-02
Fix KO Chance Text

Ntonio36:
Let the calc get the proper KO chance text for the defender's current HP, opposing to its max HP as it was previously.


Ntonio36:
@AustinXII @Zarel what about this PR?

AustinXII:
I’m not merging anything without testing it fully myself, I was going to do that tomorrow sorry for the delay

AustinXII:
thanks!

PR #238, opened 2018-03-05
Fix crashes in One-vs-All due to Metronome controls

asgdf:
The first issue was caused by the structure of the poke-info box being different between 1vsAll and 1vs1 mode, so hardcoding the DOM traversal with the `parent() `calls failed in the former case. I'm not sure if there was a reason for the roundabout implementation of "find poke-info, get id from poke-info, create new selector with that id", but this should be more tolerant to changes in the structure.

The second issue was because 1vsAll mode does not have the metronome hit counter elements (someone should add those I suppose), so it threw a TypeError due to property access of a null. Using `:visible` should not have that issue.

### DIFF
--- a/js/shared_controls.js
+++ b/js/shared_controls.js
@@ -307,11 +307,11 @@ $(".move-selector").change(function () {
 
 $(".item").change(function () {
 	var itemName = $(this).val();
-	var correspondingPokemon = $(this).parent().parent().parent().prop("id");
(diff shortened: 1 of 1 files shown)


AustinXII:
thanks and [done](https://github.com/Zarel/honko-damagecalc/commit/7a798f5b8a4023ff70de9289050153a5b73934f1)

PR #239, opened 2018-03-19
Add health bar

Ntonio36:
Adds a little cute health bar that displays how would HP appear, according to the health percent / health amount input.

The only thing I'm not quite sure of would be the color used to the "lower than the half amount" yellow, as well as from which percentage does the bar start turning red.

### DIFF
--- a/honkalculate.template.html
+++ b/honkalculate.template.html
@@ -408,7 +408,7 @@
                         <input class="percent-hp" value="100" />%)
                         <br />
                         <br />
-                        Health <div class="healthbar" style="background-color:green"></div>
+                        Health <div class="hpbar"></div>
                     </div>
                     <div class="move1">
                         <select class="move-selector small-select"></select>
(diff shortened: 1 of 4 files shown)


Zarel:
Neat! Do you have a screenshot?

Zarel:
<20% is red
<50% is yellow

Ntonio36:
percent
The bar emptied a little bit, but we're still above 50% so the green is still pretty much present.

absolute
You can see the bar adapted itself to the fact of the health being lower than the half, although the yellow is a bit too flashy, I guess.

red

And here is the dying Pokémon on a critical health, when lower than 20% HP.

Zarel:
PS already has health-bar CSS that doesn't look bad. Why don't you take those?

Ntonio36:
I'm not quite sure if the roundy aspect of the health bar fits the calc... But I'll look further on it near the weekend for sure.

Ntonio36:
In the end, I'm okay for taking the PS CSS for the health bar.. but now I need the time to do it. :joy: 

It should be done by the end of the week, hopefully.

Zarel:
lol, that's not remotely a problematic mistake to make, re: estimating times

AustinXII:
@Ntonio36 for future reference ``> =/= >=``

PR #240, opened 2018-03-27
Update calculator webpage credits and links

AustinXII:


AustinXII:
(PR'd so zarel would get notified of the change and can revert if for some reason he doesn't approve of it)

Zarel:
I'm not _very_ happy with this phrasing, but I don't really know how to improve it, so I guess it can stay.

AustinXII:
Which phrasing specifically? We could probably brain storm better phrasing if need be

Zarel:
Also your indentation is bad. :[

PR #241, opened 2018-04-01
Add Rivalry ability

Ntonio36:


Zarel:
Did you just add a gender selector to both sides?

Honestly, the main thing that needs to be fixed in the damagecalc is UX.

Ntonio36:
I reckon I wasn't quite inspired on the fact of adding (yet) a(nother) selector, but there was probably no other way around... unless there are some tricks or other stuff I don't know about yet.

js/data/pokedex.js
```diff
@@ -2110,8 +2110,6 @@ var POKEDEX_RBY = {
 };
 
 var POKEDEX_GSC = $.extend(true, {}, POKEDEX_RBY, {
-	"Magnemite": {"t2": "Steel"},
```
AustinXII: I like having gen changes at the top of the array for convenience

AustinXII:
@Zarel 100% true, maybe someone with more free time is willing to update the calc design, perhaps ask some of the “new” devs. A revamp in general would be superb, the old look was fine for 2014, but extremely boring for 2018

PR #243, opened 2018-04-05
Make hpbar rounding global

mk360:


mk360:
Messed up the accounts goddamnit D:

PR #247, opened 2018-05-01
Add HP EVs to exported sets

LegoFigure11:


js/moveset_import.js
```diff
@@ -20,6 +20,10 @@ function ExportPokemon(pokeInfo) {
 	if (gen > 2) {
 		finalText += "EVs: ";
 		var EVs_Array = [];
+		if (pokemon.HPEVs != -1 && pokemon.HPEVs > 0) { // Do HP EVs exist and are they greater than 0?
```
AustinXII: ? Does ``if (pokemon.HPEVs && pokemon.HPEVs > 0)`` not work?

LegoFigure11: It does, I just forgot it could be written that way...

PR #249, opened 2018-05-23
Update README.md

Ntonio36:


Zarel:
This does not appear to be an improvement?

Ntonio36:
Just a little `http` to `https` "typo", even if the `http` is still redirecting to the secure equivalent..

Zarel:
I would leave the URL in the text.

PR #250, opened 2018-05-28
changed Neuroforce modifier

penpexgit:
It has been discovered some weeks ago that the modifier is not 1.2 but 1.25 (so 0x1400 = 5120, and 5120/4096=1.25).


AustinXII:
Source?

penpexgit:
From Marty on Smogon USUM Battle Mechanics Research Thread [here](https://www.smogon.com/forums/threads/ultra-sun-ultra-moon-battle-mechanics-research-read-post-2.3620030/page-9#post-7765039), quoting a Japanese researcher. It was implemented on Showdown immediately after but not here.

PR #258, opened 2018-06-29
Fix Struggle GSC calculation

Ntonio36:
According to Bulbapedia:
1) Struggle is treated as a Normal-type move in Gen I
2) It becomes "typeless" from gen 2 ownards.

We can deduce from 1) that:
- Struggle inherits all Normal-type properties including:
-- The offense/defense relationship with other types ;
-- The type category of Normal, which is Physical (it was a thing before gen IV).

### DIFF
--- a/js/data/move_data.js
+++ b/js/data/move_data.js
@@ -362,6 +362,7 @@ var MOVES_RBY = {
 	'Struggle': {
 		bp: 50,
 		type: 'Normal',
+		category: 'Physical',
 		hasRecoil: 50
 	},
 	'Submission': {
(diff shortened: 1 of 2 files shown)


AustinXII:
@Marty-D

Marty-D:
All you need to do is add the Physical category for None. Everything else you listed is handled by Struggle's type in Gen 1, which is Normal, so it behaves like any other Normal move already.

Ntonio36:
@Marty-D good now ?

Marty-D:
\>All you need to do is add the Physical category for None.
No.

Ntonio36:
This time should be good.

Marty-D:
You didn't need to add Physical to Struggle, but category handling is really inconsistent so why not.

PR #261, opened 2018-07-04
Add canEvolve where reported as missing

Ntonio36:


Ntonio36:
Dunno why did it call me out for a conflict..
One more fix to go, though.

Merritty:
Only reason I inserted canEvolve before gender for Poipole was to be consistent with Klink and Cosmog who have it in that order.

AustinXII:
I noticed this in merritty’s PR but yes it’s supposed to come after, unless I’m remembering what was added first

PR #267, opened 2018-09-14
Update setdex_sm.js

DrapionSwing:
I've updated the CAP sets to make them more relevant to the metagame


AustinXII:
Just fix Travis and this is good to go

AustinXII:
Thanks! Also yay Travis is already helping :D

PR #270, opened 2018-11-19
Fix Photon Geyser damage type for equal Atk/SpA

asgdf:
It's only physical if Atk > SpA, not if Atk >= SpA.


Zarel:
Do you have a source for this?

Zarel:
Oh nvm this is the damagecalc.

PR #273, opened 2018-12-11
Update setdex

Honko:
Some CAP analyses are now on the main Smogon dex, so those CAPs now have their sets ripped and are mixed in with the rest of the Pokemon. The CAPs still at the bottom of the list don't have analyses yet and can still be manually updated.


Zarel:
Can you fix the Travis error?

Honko:
Sorry about that. Fixed.

Issue #254, opened 2018-06-21
Choice items are not taken into account in multi-hit moves after the first move


penpexgit:
Following a discussion on Smogon's Forums bug report thread for the calc [here](https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/page-4#post-7820345), the issue is the following:
Choice items (I tested both) are not taken into account in multi-hit moves (I tested Draco Meteor and Superpower for instance) after the first move.
Here is the proof:
252+ SpA Choice Specs Kingdra Draco Meteor vs. 248 HP / 208+ SpD Mantine: 165-195 (44.2 - 52.2%)
so the min damage of one hit is 165.
Now, the min damage of one hit starting at -2 SpA (to mimick what should be the second hit) is 84:
-2 252+ SpA Choice Specs Kingdra Draco Meteor vs. 248 HP / 208+ SpD Mantine: 84-99 (22.5 - 26.5%)
So the total min damage of 'Twice' Draco Meteor should be 249.
But currently we have a min of 220:
252+ SpA Choice Specs Kingdra Draco Meteor over 2 turns vs. 248 HP / 208+ SpD Mantine: 220-261 (58.9 - 69.9%)
But actually 220-165=55 which is the result of no choice Specs for the second hit:
-2 252+ SpA Kingdra Draco Meteor vs. 248 HP / 208+ SpD Mantine: 55-66 (14.7 - 17.6%)
demonstrating that indeed after the first hit, the Choice Specs is not taken into account anymore.

PR #274, opened 2018-12-31
Issue #254: Allow ability / item modifiers to affect moves used 2+ times

cjmckee:
Simple fix for this issue: https://github.com/Zarel/honko-damagecalc/issues/254

Previously, on second hits and beyond, the calculator would apply the drop modifier to the base stats, which causes it to ignore items and abilities that are affecting the attack stat such as Choice Specs. Anything that was not included in the finalMod calculations was ignored.

I replaced the raw attack with the modified attack to fix this issue.


Zarel:
Thanks!

PR #277, opened 2019-01-19
Fix Guardian of Alola damage through Protect

asgdf:
It should be doing 3/4 * 1/4, not 1/4. I copied the damage calculation straight from the server code, so the rounding should be correct.


asgdf:
Also added the "does full damage through Protect when used without a Z crystal" bit, as well as the warning message about this (the one further down the function is never shown because we return early).
Shoutouts to qtrx who pointed this out to me.

PR #292, opened 2019-03-22
Fix automatic selection of correct alt forme

asgdf:
I don't know why we bothered with all that complex logic when we literally have the correct forme name available from the user's selection.


AustinXII:
Awesome thank you, the reason for the complex logic was because it was just copied over when I was creating the alt forme selection

AustinXII:
This should theoretically close #291

PR #295, opened 2019-04-08
Both to all

WheatiesTM:
this is just a small change that allows the Import set function to work when there are more than two poke-info elements on the html page. This shouldn't affect the function of the calculator currently but allows my planned type calculator to work with more than two pokemon.


WheatiesTM:
not quite sure why travis is failing it because the functions it says are not defined were not defined within that file on the previous build I believe, will look more into it

AustinXII:
Just add the functions into the global section into the eslinttrc file

PR #297, opened 2019-04-09
Both to all i++ fix

WheatiesTM:
fixed a bug in the Both to All PR


WheatiesTM:
@AustinXII fixed, will be more careful next time

PR #307, opened 2019-05-01
Make Metronome item more accurate

wartab:
This updates how the Metronome damage boost is applied according to this:
https://www.smogon.com/bw/articles/bw_complete_damage_formula

I removed the feature that attempts to calculate the damage over an estimate of turns as
1) it did not work correctly before
2) it is too complex too implement
3) I believe seeing the damage for a given turn is much more useful (you might be using the move on a different Pokémon after a few turns, stat boosts/drops may have happened inbetween turns, etc.)

I also made this independent of stat-dropping moves.

### DIFF
--- a/js/damage.js
+++ b/js/damage.js
@@ -725,6 +725,15 @@ function getDamageResult(attacker, defender, move, field) {
 		finalMods.push(0xC00);
 		description.defenderAbility = defender.ability;
 	}
+	if (attacker.hasItem("Metronome") && move.metronomeCount >= 1) {
+		var metronomeCount = Math.floor(move.metronomeCount);
+		if (metronomeCount <= 4) {
+			finalMods.push(0x1000 + metronomeCount * 0x333);
+		} else {
+			finalMods.push(0x2000);
+		}
+		description.attackerItem = attacker.item;
+	}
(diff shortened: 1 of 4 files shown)


js/damage.js
```diff
@@ -725,6 +725,15 @@ function getDamageResult(attacker, defender, move, field) {
 		finalMods.push(0xC00);
 		description.defenderAbility = defender.ability;
 	}
+	if (attacker.hasItem("Metronome") && move.metronomeCount >= 1) {
+		var metronomeCount = Math.floor(move.metronomeCount);
+		if (metronomeCount <= 4) {
+			finalMods.push(0x1000 + metronomeCount * 0x333);
```
AustinXII: Heh I was right, I told Antonio there was a much easier way to do this without looping

AustinXII:
Thanks!

js/damage.js
wartab: I think that was he was trying to implement was different though, after updating what he did. Think he tried to add up the damage from every turn leading to the current turn.

AustinXII: Yeah ik but his approach was still wrong

PR #314, opened 2019-05-07
Fix Mega pokemon getting their base form ability

AustinXII:


AustinXII:
@Zarel here's a fix but sadly it doesn't account for pokemon with two mega formes, I think it will require some more hardcoding :/

js/moveset_import.js
```diff
@@ -262,7 +260,11 @@ function addSets(pokes) {
 					currentPoke.nameProp = "Custom Set";
 				}
 				currentPoke.isCustomSet = true;
-				currentPoke.ability = getAbility(rows[i + 1].split(":"));
+				if (currentPoke.name.indexOf("-Mega") !== -1) {
+					currentPoke.ability = getAbility(rows[i + 1].split(":"));
+				} else {
```
AustinXII: will probably need something like ``currentPoke.name.indexOf("-Mega-X) {currentPoke.ability = POKEDEX_SM[currentPoke.name].ability;}`` and ``currentPoke.name.indexOf("-Mega-Y") {currentPoke.ability = POKEDEX_SM[currentPoke.name].altAbility;``

js/moveset_import.js
```diff
@@ -262,7 +260,11 @@ function addSets(pokes) {
 					currentPoke.nameProp = "Custom Set";
 				}
 				currentPoke.isCustomSet = true;
-				currentPoke.ability = getAbility(rows[i + 1].split(":"));
+				if (currentPoke.name.indexOf("-Mega") !== -1) {
```
Zarel: It's interesting that you're checking for non-mega first, and then leaving mega for the `else`. I would do it the other way around: the default should be non-mega.

And also, you should probably check that `POKEDEX_SM[currentPoke.name]` and `POKEDEX_SM[currentPoke.name].ability ` even exist.

AustinXII: Honestly I just did this fast so I wouldn’t forget how I did it tomorrow lol. But yes that is probably better, ty!

js/moveset_import.js
```diff
@@ -262,7 +260,12 @@ function addSets(pokes) {
 					currentPoke.nameProp = "Custom Set";
 				}
 				currentPoke.isCustomSet = true;
-				currentPoke.ability = getAbility(rows[i + 1].split(":"));
+				if (currentPoke.name.indexOf("ite") === -1) {
```
Zarel: Err, either `currentPoke.name` is named very badly, or this is wrong and you should be testing item name instead.

AustinXII: Yeah this should be ``currentPoke.item`` but it’s not working as intended so I’m brainstorming a bit more.

PR #318, opened 2019-05-12
Support importing custom tiers

Ortheore:
Hey, I added the option to import custom tiers into 1vAll and Allv1 modes. Currently it only accepts importables. Might need some polishing. At this point I'm no longer interested in developing it further, but I thought I'd share it with you anyway in case you were interested. If not, all good.


AustinXII:
I don’t think this is something we need, right? @Zarel

Zarel:
No clue what this is and how needed it is.

AustinXII:
Cool, I know what it is but I think it’s not needed. If @Ortheore comes up with a good reason for it’s implementation I’ll reopen it

PR #320, opened 2019-06-07
Bug fix- KO text in all vs one

Ortheore:
Hi it's not an esoteric feature of questionable usefulness this time.

Fixed issue where KO text would not display correctly in all vs one.

Still some work to be done to account for stat drop moves and metronome, since all I did was set usedTimes and metronomeCount to default values if undefined, which they always are in all vs one due to lack of UI for attackers


AustinXII:
Can you fix Travis? It’s just your spacing on if statements that’s upsetting it

Ortheore:
Adjusted my formatting to be consistent with rest of code. Seems to have passed?

AustinXII:
Thanks!

PR #322, opened 2019-07-08
Reorganize view and logic code in preparation for making a 'damage' subpackage

scheibo:
_And so it begins..._

**Context:** [pkmn.cc/damagecalc-revamp](https://pkmn.cc/damagecalc-revamp)

No guarantee files won't move around further, but I had to start somewhere. This PR should introduce no functional changes, but given that there's no tests (yet) its hard to guarantee it won't break things (though it works fine in my browser). This will probably be the most dramatic PR in terms of files moving around and structure changing, hopefully the changes later on will be more localized.

My first milestone will be splitting `damage/` into an independent module (with its own `package.json`) that can be build and compiled separately (to then be rewritten into Typescript). I'm deliberately punting on Webpack and the complexity it incurs with my silly 'copy the files to `dist/`' build scheme for the time being until the view code is more untangled and there are less globals etc to worry about.

_You probably want to merge these commits individually as opposed to squash-merging - both should pass 'tests' individually and will make more sense in the commit history as distinct commits._


AustinXII:
You’re honestly such a legend dude. Icr if you’ve addressed this is your to-do for the calc but I’d imagine select2 needs to be updated after all these years if you want to tackle that

Zarel:
> In an ideal world, we would be able to share the same logic around damage calculation between the simulator and the calculator, but this does not seem like a realistic option due to the fact that the simulator is concerned with considerably more than just the damage component of moves (accuracy, effects, etc), how the PRNG is used (the simulator would need to change how and where it makes PRNG calls to be usable by the non-stochastic calculator) and how the simulator code is structured in general.

There'd probably be nonzero code duplication, but I think at least the `onDamage` etc functions in PS's data files can be shared.

If not, we should probably introduce `onDamageCalc` functions into PS's data files, so they're all in one place.

Zarel:
Also, if Honko has code to import Smogon sets, that should probably also be integrated with PS's data. It would be quite useful for PS's teambuilder, for instance.

AustinXII:
@Honko

scheibo:
> Icr if you’ve addressed this is your to-do for the calc but I’d imagine select2 needs to be updated after all these years if you want to tackle that

Somewhat OOS, other than the fact that if I do manage to get Webpack working I'd just list the select2 dep in the `package.json` instead of vendoring it.

> There'd probably be nonzero code duplication, but I think at least the `onDamage` etc functions in PS's data files can be shared.
> 
> If not, we should probably introduce `onDamageCalc` functions into PS's data files, so they're all in one place.

I would like to [carve up PS into modular chunks](pkmn.cc/modular-ps) and share the data files between PS proper and the damage calc (Stage 2 of this revamp). Sharing `onDamage` etc functions is not currently something I have in mind, but would potentially be feasible further down the line. Just sharing the raw data (base power, types, categories etc) does not preclude eventually sharing more, it just isn't something I could figure out how to do cleanly at this time.

> Also, if Honko has code to import Smogon sets, that should probably also be integrated with PS's data. It would be quite useful for PS's teambuilder, for instance.

The code to import sets is pretty trivial? See https://github.com/pkmn-cc/EPOke/blob/master/src/analyses.ts for example. And [others have already integrated this information with PS's team builder through extensions](https://www.smogon.com/forums/threads/setdeli-smogon-sets-in-the-ps-teambuilder.3652231/). If this is something you're interested in it would be relatively trivial to do, https://pkmn.cc/smogon-ps-integration for some considerations.


CC @jaredgrady @Nineage who both seem to have been working on Technical Projects in this area and may be interested in contributing here (given this revamp would render their [Damage Calc API](https://www.smogon.com/forums/threads/damage-calculator-api.3599759/) project obsolete and they're the ones behind the Smogon Sets + PS plugin)

AustinXII:
Fender and nineage, help a man out

jaredgrady:
@scheibo which aspect of the integration are you looking for contributions on? I'd be happy to assist with building Smogon set retrieval directly into the PS client if there is interest in that. I think the long term scalability issue that the plugin we wrote has is its reliance on web scraping to retrieve the sets. In an ideal world, Smogon would provide an endpoint from which we could retrieve that information in a workable and consistent format, rather than us having to maintain data files for analysis that are often changing regularly.

scheibo:
@jaredgrady - You an @Nineage seem to be doing good work and I was hoping to encourage you two to consider contributing some of it upstream where it can have the most impact :). 

From the comments on this PR it seems like @Zarel would be interested - maybe he can comment on what he had in mind integration wise. I agree the scraping is fragile, and doing something like https://github.com/pkmn-cc/EPOke/blob/master/src/analyses.ts#L73 to pull out the JSON blob is pretty silly when we should be able to just hit an endpoint (and actually, we don't really care about analysis notes for this integration I'd imagine, we probably only want the raw set data). I don't know if summoning @monsanto on GH works, but I can ping chaos elsewhere to ask if theres a better way to do it.

As I briefly touch upon in the [integration proposal 1-pager](https://pkmn.cc/smogon-ps-integration), I'm not positive if we'd want to be querying Smogon endpoints live as opposed to pre-building a index to serve from `play.pokemonshowdown.com` along with our other data files. I'd suggest [opening an issue on the client](https://github.com/Zarel/Pokemon-Showdown-Client/issues/new) to discuss this further :)

Honko:
Re: Importing Smogon sets, for the record I scrape them essentially the same way EPOke does. If a better way exists, nobody ever told me about it.

I'm happy to see someone finally modernize the calc. Austin's done an admirable job of keeping this thing afloat for the past few years, but I didn't know anything about modern web development when I wrote it in ~2012, it hasn't been meaningfully revamped since then, and the result is pretty clearly a maintenance nightmare at this point.

Two comments about the proposed changes that you can take for whatever they're worth:

1) >‘Too many knobs’. While being able to support every possible interaction is very valuable, the UI is not streamlined for the standard use case - options that are infrequently used (eg. ‘Battery’) can and should be collapsed and require extra effort to access.

I 100% agree there are too many knobs now and that it's unsustainable to keep adding more without collapsing the lesser used ones. Just be careful about where you draw the line for what's immediately accessible. The main reason I built this calculator in the first place (and, I strongly suspect, the reason it became the standard one) is because all the other calculators at the time required too many clicks to get the answer you wanted. [DJD's calc](https://www.smogon.com/calc/) was a lot nicer than mine, but it took 6/8/10 clicks to get a result that you can get in 2/3/4 clicks on mine. People want their calcs fast. Obviously once you get to the point of integrating with the PS client this becomes moot, but it's something to keep in mind as you make changes in the meantime.

2) > Processed usage statistics (which implements some ideas for handling spread reports as suggested on Smogon) can be used to come up with median stat spreads instead of relying on manually curated sets/sets pulled from Smogon, based on the idea that when performing calculations, calculating against a legal set is only absolutely required in niche cases.

I'm not sure I agree with the premise. It's probably true for the average PS user, but I suspect the Smogon tournament playerbase would be strongly opposed to moving away from curated sets, if that's a user group you care about. The calc currently includes usage-based sets as filler for Pokemon that don't have a Smogon analysis, and I remember seeing quite a few complaints about how bad those sets tend to be.

That all said, your proposal looks great to me and I look forward to seeing it happen. :+1:

scheibo:
> I'm happy to see someone finally modernize the calc. Austin's done an admirable job of keeping this thing afloat for the past few years, but I didn't know anything about modern web development when I wrote it in ~2012

Agreed, thanks to both of you for your work! The fact that the calculator has stood around so long is testament to its original design and Austin's excellent effort maintaining it.

> I 100% agree there are too many knobs now and that it's unsustainable to keep adding more without collapsing the lesser used ones. Just be careful about where you draw the line for what's immediately accessible. The main reason I built this calculator in the first place (and, I strongly suspect, the reason it became the standard one) is because all the other calculators at the time required too many clicks to get the answer you wanted. [DJD's calc](https://www.smogon.com/calc/) was a lot nicer than mine, but it took 6/8/10 clicks to get a result that you can get in 2/3/4 clicks on mine. People want their calcs fast. Obviously once you get to the point of integrating with the PS client this becomes moot, but it's something to keep in mind as you make changes in the meantime.

Definitely. I don't know if its clear from the design document, but the main thing I'm looking to do with this revamp is:

1) modularize the logic into a reusable package
2) make the core more robust through typechecking and tests
3) do some code cleanup to make the code follow more of the patterns and best practices of the slightly more mature Pokemon-Showdown codebase
4) try to share code with the rest of the Pokemon-Showdown ecosystem

Actually adding features or changing the UI are not really things I'd actively be pursuing at this time, its more that I just figured if I was critically reviewing the code and design I might as well review the UI and features as well. I'm less confident about what my exact vision is with respect to what the UI would actually look like, and I'd be looking to collaborate with you and Zarel there if I ever had time to go down that path (there's too much code to play around with and too little time. I was hoping my doc might inspire other interested parties once I'm done with my 'clean up')

> I'm not sure I agree with the premise. It's probably true for the average PS user, but I suspect the Smogon tournament playerbase would be strongly opposed to moving away from curated sets, if that's a user group you care about. The calc currently includes usage-based sets as filler for Pokemon that don't have a Smogon analysis, and I remember seeing quite a few complaints about how bad those sets tend to be.

I think the important thing I'm trying to get across is that 'set' is very limiting in a calculator when it doesnt need to be. There's no reason we only need to show 4 moves - in fact, even the curated sets have slashes in them to indicate alternatives. I'd agree that one thing processed usage stats doesn't cover is providing a good picture of the _gestalt_, ie which combinations are typically found together, but this could be solved by simply highlighting the combinations that go together by joining the curated sets with statistics.

This feature is fairly pie in the sky, and could really benefit from a working demo to better outline what my vision actually is here. However, similarly to the UI changes above, I don't really plan on attempting to tackle this anytime soon - my immediate goals are purely under the covers and developer-visible, not end user-visible. :)

PR #323, opened 2019-07-09
Unify damage calculation entry point

scheibo:
This change starts to define what the API exposed from the `damage` subpackage is meant to look like (ie. as [demonstrated by the strawman prototype](https://github.com/pkmn-cc/dmg/blob/master/calc.ts)).

In order to unify the API, this **change introduces a performance regression** where stats for the attacker and defender are computed each move (i.e. 4x instead of 1x). Computing the stats is not super performance intensive, and I don't believe avoiding this duplicated work is really worth the extra complexity. I plan to do a performance optimization pass as part of this revamp which I would expect to buy back more time than the regression introduced by this change, but if this regression is unacceptable I will add to the API surface area to allow for the existing behavior to be preserved.

The next PR will continue to chip away at the globals and work towards encapsulating the `damage` subpackage further.


Zarel:
I accidentally merged this with the commit summary 'Calc damange'.

I assume that was a mistake on your part. :p

scheibo:
ROFL. Yes, sorry. "Unify damage calculation entry point" was the intended title here.

AustinXII:
Just curious, are you comparing current calc damage outputs with your updated calc damage outputs to ensure nothing is going awry?

scheibo:
> Just curious, are you comparing current calc damage outputs with your updated calc damage outputs to ensure nothing is going awry?

Yes. I just have a window open with the current calc and a window open with my local build and I toggle through all the modes and gens and maybe tweak a few buttons along the way. I added a bug with yesterdays PR that I fixed in https://github.com/Zarel/honko-damagecalc/commit/835db8c27b6a74203827699efe11193984b08e37, but this one is bug-for-bug-compatible, to the best of my manual testing*


*GSC seems bugged at `HEAD`? I see `NaN%` whenI toggle through it in `One-vs-One` mode, and `(No Move)` in some of the Honkalculate views. Known issue? Either way, my refactoring doesn't fix or break this further from what I can tell

AustinXII:
> *GSC seems bugged at `HEAD`? I see `NaN%` whenI toggle through it in `One-vs-One` mode, and `(No Move)` in some of the Honkalculate views. Known issue? Either way, my refactoring doesn't fix or break this further from what I can tell

Known just haven’t had time to fix

AustinXII:
Awesome, I glad you’re testing 👍

PR #324, opened 2019-07-11
Make calc/ an actual subpackage, begin converting to Typescript

scheibo:
- currently the package is not properly usable in environments with
modules and expects to be required by the browser. Going forward,
once everything has been converted to Typescript, `index.ts` will
`export` all the types for Node, but the calculator UI can still
simply rely on globals (provided it sets up `require` and `exports`
properly, as done in this change). Really, a bundling solution like
Webpack would be desirable to avoid the preamble setup, but its the
simplest solution with the least moving parts, and isn't *so*
fragile because while `calc/` is a self-contained subpackage, it
still lives as a directory under the UI and we can easily adjust the
UI setup logic in the same commit.
- the final published package name should probably be something like
@pokemon-showdown/calc or @psim/calc, but since its not actually
usable as a package in its current state its not worth arguing over
naming at this point.
- data/nature.ts and stats.ts were converted as a proof of concept,
followup changes will do more conversion.
- I want to set up automatic code formatting for this subpackage
with `google/gts` (which uses tslint and prettier), but
unfortunately it also formats the unmigrated Javascript in the
subpackage with no way to get it to ignore the files, so I'll wait
to introduce a code formatter once everything has been migrated. For
now, I've copied the tslint.json that it would use for linting to
minimize any churn once `gts` starts being used.


calc/package.json
```diff
@@ -0,0 +1,42 @@
...
+  },
+  "author": {
+    "name": "Guangcong Luo",
+    "email": "zarel@zarel.com",
+    "url": "http://guangcongluo.com"
+  },
+  "contributors": [
+    {
+      "name": "Mathieu Dias-Martins",
+      "email": "marty@hisemail.com"
```
AustinXII: We can remove this, i don’t think Marty cares enough about the calc to want to be notified about Travis, etc

AustinXII:
Easy enough Travis fix when you get around to it. Besides that this looks good!

calc/package.json
AustinXII: Also it’s obviously not a real email, I was going to ask him about adding it but figured it wouldn’t matter

scheibo:
```
[17:57] Zarel: yeah, no objections
```

OK, well both of you have signed off on the approach. I'll fix whatever the heck Travis is complaining about (it works finally locally!) and push tomorrow after further sanity testing, just wanted to have some more eyes on this one given its going to determine the future path I take.

Will probably convert most of the `calc/` to Typescript next - likely not tomorrow, but mostly likely before Sunday. Goal is to have a typed-module friendly library usable by the end of the week.

calc/.eslintrc
```diff
@@ -0,0 +1,343 @@
...
+        "no-case-declarations": "off",
+        "no-div-regex": "error",
+        "no-else-return": "off",
+        "no-labels": [
+            "error",
+            {
+                "allowLoop": true,
+                "allowSwitch": true
+            }
+        ],
```
AustinXII: We should probably go through these and remove any un-needed ones as this was copied from PS

scheibo: Well, this is only temporary anyway. Once I've converted to TS I will be deleting it in favor of the `tslint.json` (and automatic code formatting).

scheibo: I actually realized I could just remove `"root": true` and inherit almost all of this. So while its still temporary, its now short as well.

AustinXII:
@scheibo grats on collaborator :). Would you like to be put in the contributors section in the package file? @Honko you should be added to if you’d like

scheibo:
Maybe once I've done more than rearrange deck chairs and actually contributed something substantive

AustinXII:
Fair enough @scheibo, I just believe you enough that you will get more than that done but it’s your call. Once you’re fully done your calc plans you definitely will deserve it

PR #337, opened 2019-08-04
Fix Mold Breaker in Gen 5-7

### LINKED ISSUE #308: Mold Breaker is not being calculated
Mold breaker is not affecting damage. For example, mold breaker haxorus at +2 attack does the same to unaware clefable as an unboosted haxorus

scheibo:
Fixes #308. Austin - opened this as a PR instead of just pushing because I'm not sure why there's a `defAbility` variable... `gen4.ts` just sets `defender.ability = ''`, so that's what I did here as well.

### DIFF
--- a/calc/src/mechanics/gen7.ts
+++ b/calc/src/mechanics/gen7.ts
@@ -95,7 +95,6 @@ function calculateModern(
     damage.push(0);
     return result;
   }
-  let defAbility = defender.ability;
   const defenderIgnoresAbility = defender.hasAbility(
     'Full Metal Body',
     'Prism Armor',
@@ -113,12 +112,12 @@ function calculateModern(
     ].indexOf(move.name) !== -1;
   if (!defenderIgnoresAbility) {
     if (attackerIgnoresAbility) {
-      defAbility = '';
+      defender.ability = '';
       description.attackerAbility = attacker.ability;
     }
   }
   if (moveIgnoresAbility) {
-    defAbility = '';
+    defender.ability = '';
   }
 
   const isCritical =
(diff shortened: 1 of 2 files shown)


AustinXII:
``asgdf on Apr 24  Contributor
Noo you're not working with a copy of the pokemon object, you're modifying it for all following damage calculations``
I did this a while ago and it would make it so every ability would ignore other abilities and not just mold breaker and bunch

AustinXII:
Defability is so the statement only gets ran once

scheibo:
I'm not following, but `defender` in this method *is* in fact now a copy (https://github.com/Zarel/honko-damagecalc/blob/master/calc/src/calc.ts#L33-L36), so I'm not sure @asgdf's comment still applies?

AustinXII:
Ah, that was the reasoning it didn’t work before and was done like that. if you tested this and it works, it should be good to go 👍

AustinXII:
What about mold breaker like moves?

scheibo:
> What about mold breaker like moves?

Not aware of any of those? https://pokemondb.net/pokebase/269287/what-moves-abilities-can-change-enemys-ability-ignore-items

Zarel:
@scheibo Your link seems to be from before Gen 7. Gen 7 has a few ability-bypassing moves; namely Sunsteel Strike, Moongeist Beam, Photon Geyser, and Light That Burns the Sky.

scheibo:
Yeah, Austin pinged me on Discord. https://github.com/Zarel/honko-damagecalc/blob/master/calc/src/mechanics/gen7.ts#L104 covers thoses

Issue #332, opened 2019-08-04
Move repository to github.com/Pokemon-Showdown/Pokemon-Showdown-Calc


scheibo:
Afterwards, we can cut a v0.0.1 release of `@pokemon-showdown/calc`

scheibo:
Also, per discussion, we will host the page at `calc.pokemonshowdown.com` with `damagecalc.pokemonshowdown.com` and `pokemonshowdown.com/damagecalc` redirecting to the new subdomain.

scheibo:
> Also, per discussion, we will host the page at `calc.pokemonshowdown.com` with `damagecalc.pokemonshowdown.com` and `pokemonshowdown.com/damagecalc` redirecting to the new subdomain.

I did this part, only @Zarel can perform the repo move.

PR #341, opened 2019-08-11
Add Google Analytics tracking code

### LINKED ISSUE #329: Add analytics around page views and which modes are being used
- Set up a new property for the damage calc in Google Analytics
- add something like following snippet:

```html
<!-- Google Analytics -->
<script>
(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
})(window,document,'script','//www.google-analytics.com/analytics.js','ga');
ga('create', 'UA-XXXXXX-X', 'pokemonshowdown.com/damagecalc');
ga('send', 'pageview');
</script>
<!-- End Google Analytics -->
<script>
	window.onerror = function(err, uri, line) {
		var msg = uri + ':' + line + ': ' + err;
		if ('ga' in window) {
			ga('send', 'event', 'Client error', msg);
		} else {
			alert('ERROR: ' + msg);
		}
		return false;
	};
</script>
```

Though we also want to track:
  -  `generation`
  - `mode` (one vs one, one vs all, all vs one)
  - `format` (standard, vgc, random, bss, etc)

scheibo:
Fixes #329.

### DIFF
--- a/src/honkalculate.template.html
+++ b/src/honkalculate.template.html
@@ -9,6 +9,24 @@
     <link type="text/css" rel="stylesheet" href="https://cdn.datatables.net/colvis/1.1.1/css/dataTables.colVis.css" />
     <link type="text/css" rel="stylesheet" href="./js/vendor/select2/select2.css" />
     <link type="text/css" rel="stylesheet" href="./css/main.css?" />
+    <!-- Google Analytics -->
+    <script>
+        (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+        (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+        m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+        })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
+
+        ga('create', 'UA-XXXXXX-Y', 'calc.pokemonshowdown.com');
+        ga('send', 'pageview');
+
+        window.onerror = function(err, uri, line) {
+            if ('ga' in window) {
+                var msg = uri + ':' + line + ': ' + err;
+                ga('send', 'event', 'Damage calculator error', msg);
+            }
+            return false;
+        };
+    </script>
 </head>
 
 <body>
--- a/src/honkalculate.template.html
+++ b/src/honkalculate.template.html
@@ -9,23 +9,14 @@
     <link type="text/css" rel="stylesheet" href="https://cdn.datatables.net/colvis/1.1.1/css/dataTables.colVis.css" />
     <link type="text/css" rel="stylesheet" href="./js/vendor/select2/select2.css" />
     <link type="text/css" rel="stylesheet" href="./css/main.css?" />
-    <!-- Google Analytics -->
+    <!-- Global site tag (gtag.js) - Google Analytics -->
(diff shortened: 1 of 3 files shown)


src/honkalculate.template.html
```diff
@@ -9,6 +9,24 @@
...
     <link type="text/css" rel="stylesheet" href="./js/vendor/select2/select2.css" />
     <link type="text/css" rel="stylesheet" href="./css/main.css?" />
+    <!-- Google Analytics -->
+    <script>
+        (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+        (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+        m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+        })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
+
+        ga('create', 'UA-XXXXXX-Y', 'calc.pokemonshowdown.com');
```
scheibo: @Zarel - please set up the property for this.

scheibo:
We should be able to get `mode` from the `pageview` because its a query param, but I've added an event to track `gen`. Though we won't be getting both `gen` and `mode` linked, so maybe we should change that?

src/honkalculate.template.html
Zarel: We got clogged up on whether or not this should be a subdomain or subdirectory. I guess separate property is fine, though.

Zarel:
I set up the property. Analytics tells you to use this code:

```html
<!-- Global site tag (gtag.js) - Google Analytics -->
<script async src="https://www.googletagmanager.com/gtag/js?id=UA-26211653-3"></script>
<script>
  window.dataLayer = window.dataLayer || [];
  function gtag(){dataLayer.push(arguments);}
  gtag('js', new Date());

  gtag('config', 'UA-26211653-3');
</script>
```

src/honkalculate.template.html
scheibo: You had [agreed to calc.pokemonshowdown.com](https://discordapp.com/channels/@me/551083939561865238/607642014280056839), and the rename issue (#332) assigned to you covers moving to a subdomain.

Zarel: Yes, that's why separate property is ifne.

scheibo:
> Analytics tells you to use this code:

Fine, whatever, we'll use the new hotness.

I removed the `gen` stuff because I want to switch to `calc.pokemonshowdown.com/gen1` etc as discussed on Discord and i'll add tracking for that then (or maybe we'd then get it simply through the pageviews metric)

AustinXII:
:+1 now we wait

Zarel:
Are we not listening for client errors, then?

scheibo:
It's supposed to be automatic with google tag manager - you just do setup in the UI AFAICT.

Zarel:
Can you do the setup? (You should have access now.)

Zarel:
> It's supposed to be automatic with google tag manager - you just do setup in the UI AFAICT.

https://developers.google.com/analytics/devguides/collection/gtagjs/exceptions

It does not appear to be automatic. Am I missing something?

scheibo:
I confused Google Tag Manager a www.googletagmanager.com/gtag/js. I wonder how?

PR #344, opened 2019-08-12
Add support for Shell Bell

scheibo:
This mostly seems to work, but it causes the UI to deform in edge cases. Not sure about whether we care or whether there's a good solution:

Screen Shot 2019-08-12 at 19 15 18

This also fixes recovery behavior to cap properly - recovery should be capped by the defender's HP stat (test values changed to reflect this).

### DIFF
--- a/calc/src/data/items.ts
+++ b/calc/src/data/items.ts
@@ -78,6 +78,7 @@ const ADV = GSC.concat([
   'Razz Berry',
   'Salac Berry',
   'Sea Incense',
+  'Shell Bell',
   'Silk Scarf',
   'Sitrus Berry',
   'Soul Dew',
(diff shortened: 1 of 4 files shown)


AustinXII:
Yeah we should try fixing the UI

scheibo:
Sick of playing around with overflow CSS. Who's even going to use fucking Shell Bell anyway? I'm pushing this because it fixes the recovery cap issues and will look fine in the UI as long as its not also paired with a recoil move. Can fix this in a followup

Zarel:
I would say "healed" instead of "recovered" . Probably not enough to fix the overflow issue but I think it's more readable.

scheibo:
If I go from 'recovered' -> 'healed' and 'recoil damage' -> 'recoil' then it fixes the issue for more cases, though still is subject to wrap at extremes

Zarel:
I would do that.

scheibo:
I went to go do that, but I forgot the recoil message is actually more complicated: 
https://github.com/Zarel/honko-damagecalc/blob/master/calc/src/desc.ts#L199-L220

PR #347, opened 2019-09-03
Correct Gen 3 damage calculation

scheibo:
- Doubles spread damage is 50%, not 75%
 - Only Physical attacks are guaranteed a minimum of 1 before the +2

Reference: https://web.archive.org/web/20170622160244/http://upcarchive.playker.info/0/upokecenter/content/pokemon-ruby-version-sapphire-version-and-emerald-version-timing-notes.html

### DIFF
--- a/calc/src/data/moves.ts
+++ b/calc/src/data/moves.ts
@@ -9,7 +9,7 @@ export interface MoveData {
   readonly type: Type;
   readonly category?: Category;
   readonly hasSecondaryEffect?: boolean;
-  readonly isSpread?: boolean;
+  readonly isSpread?: boolean | 'allAdjacent';
   readonly makesContact?: boolean;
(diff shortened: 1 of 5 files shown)


scheibo:
`isSpread` is not sufficient unfortunately, we need to change `isSpread` from a boolean to an enum to be able to distinguish between `adjacentFoes` and `allAdjacent` per discussions with Marty

Zarel:
`boolean | 'allAdjacent'`?

scheibo:
currently is `isSpread?: boolean` (ie/ only set when true), so `isSpread?: 'allAdjacent' | 'allAdjacentFoes'` should work.

scheibo:
Went with `boolean | 'allAdjacent'` in case something goes from a spread move to not a spread move in the future somehow.

PR #349, opened 2019-09-04
Change Prettier's bracket spacing configuration

scheibo:


Zarel:
Should we also look at other Prettier configuration?

scheibo:
Not sure I prefer this TBH. It forces you to not use spaces *everywhere*, when I want them sometimes (and yes, I know if we didnt use an automatic formatter that wouldnt be a tradeoff we'd need to discuss). I think in the previous commit they were desirable, but I think on the whole I prefer spaces (ie/ how the code looks before this change). @AustinXII  - what is your preference?

> Should we also look at other Prettier configuration?

I want to turn on consistent quotes in object literals, but thats a different commit IMO

Zarel:
Okay, the other defaults make sense to me.

Zarel:
> I want to turn on consistent quotes in object literals, but thats a different commit IMO.

I'd generally disagree, unless we're talking about different things?

scheibo:
> I'd generally disagree, unless we're talking about different things?

Another 'it depends'. But for an object with like only 10 keys and 2 have to be quoted it bugs me that the rest arent consistent. Also, remember - this is automatically enforced so people editing the file can just forget quotes and then run `npm run fix` (or forget to and have to followup with shameful commits after the fact...)

PR #352, opened 2019-10-12
Add GSC Present, fix old gen residual and status damage

Enigami:
Adds Present to GSC moves, using its unusual formula for damage.
https://www.youtube.com/watch?v=XJaQoKtrEuw

Fixed GSC Sandstorm to add 1/8 damage per turn instead of 1/16.
https://bulbapedia.bulbagarden.net/wiki/Sandstorm_(weather_condition)#Effects

Fixed RBY Poison/Burn to add 1/16 damage per turn instead of 1/8.
https://bulbapedia.bulbagarden.net/wiki/Poison_(status_condition)#Generation_I
https://bulbapedia.bulbagarden.net/wiki/Burn_(status_condition)#Generation_I


calc/src/data/moves.ts
```diff
@@ -1349,6 +1354,8 @@ const ADV: {[name: string]: MoveData} = extend(true, {}, GSC, {
   },
 });
 
+delete ADV['Present'];
```
AustinXII: Don’t delete the move

calc/src/mechanics/gen2.ts
```diff
@@ -58,7 +58,7 @@ export function calculateGSC(attacker: Pokemon, defender: Pokemon, move: Move, f
     return result;
   }
 
-  const lv = attacker.level;
+  var lv = attacker.level;
```
AustinXII: const is correct

calc/src/mechanics/gen2.ts
```diff
@@ -127,6 +127,18 @@ export function calculateGSC(attacker: Pokemon, defender: Pokemon, move: Move, f
     at = Math.floor(at / 4) % 256;
     df = Math.floor(df / 4) % 256;
   }
+  
+  // Gen 2 Present has a glitched damage calculation using the secondary types of the Pokemon for the Attacker's Level and Defender's Defense.
+  if (move.name === 'Present') {
```
AustinXII: This doesn’t seem like the right calculation? Also can you write tests?

calc/src/mechanics/gen2.ts
Enigami: Const is not correct, Present alters the level of used in the calculation. If lv is a constant, it cannot be modified which prevents Present from having an accurate calculation,

calc/src/data/moves.ts
Enigami: Present's unusal and useful properties ONLY exist in GSC. Is there any reason to keep it in the calculator in ADV onwards, when the move's damage calculation is changed and no longer useful?

calc/src/mechanics/gen2.ts
Enigami: The calculation is correct. When Present is used, after the Attack/Defense/Level values are read from and stored, they are overwritten by Present's buggy code.

The "at"/Attack value is set to 10.
The "df"/Defense value becomes the Attacker's secondary type index.
The "lv"/Level value becomes the Defender's secondary type index.
Mono-type Pokemon have the same type twice internally, so these Pokemon have a primary and secondary type of 'Normal', 'Water', etc.

This means that when used by a Normal-type, the attack has 10 Attack against a 0 Defense target (Normal's index = 0, Attacker's type = Normal, Defender's Defense is set to Attacker's type which is 0), with the type index of the Defender used in place of the Attacker's level, which is lower for 'physical'-types (Normal=0, Fighting=1, etc.) and higher for 'special'-types (Fire=20, Water=21, etc.). This makes the attack unreasonably powerful against Pokemon such as Mewtwo, Mew, Celebi, or any other Pokemon with a secondary type index of 20 or higher (any 'special' type), to the point a Level 5 Togepi has the capacity to OHKO a level 100 Umbreon as shown in Crystal_'s Youtube video if it gets a 120 BP Present (1/10 Chance).

This is what my modification to the calculator will show for Present:
Togepi (set to Level 5) Present (set to 120 BP) vs. Umbreon (set to Level 100): 370-435 (94.1 - 110.6%) -- 64.1% chance to OHKO after Leftovers recovery

This happens in Crystal_'s video at the point Togepi nearly OHKOs the Umbreon. Due to Present overwriting the Attacker's Attack/Attacker's Level/Defender's Defense used for calculation, Togepi's attack becomes a level 27 (Dark's index is 27) STAB 40/80/120 BP attack (or a 1/4 Heal) with an Attack stat of 10 against a Defense stat of 0. This overwrite happens *after* boosts, screens and so on are applied to Attacker's Attack/Defender's Defense, so Present ignores the Attack stage of the attacker, as well as the Defense stages and screens of the defender.

Using Showdown's GSC damage calculation shows that:
let damage = level[27] * 2 -- damage = 54
damage = Math.floor(damage[54] / 5) -- damage = 10
damage += 2 -- damage = 12
damage *= 120 (max damage of Present) -- damage = 1440
damage *= attack[10] -- damage = 14400
damage = Math.floor(damage[14400] / defense[0]) --
damage = clampIntRange(Math.floor(damage[14400] / 50) = 288, 1, 997) -- damage = 288
damage += 2 -- damage = 290

--STAB Bonus--
damage += Math.floor(damage / 2) -- damage = 435

435 = max damage of STAB 120BP Present from a mono Normal-type vs. Umbreon, same as my calc shows.
435 * 217 / 255 = 370, the minimum damage of STAB 120BP Present from a mono Normal-type vs. Umbreon, same as my calc shows.

calc/src/mechanics/gen2.ts
AustinXII: Than it should be ``let``

calc/src/data/moves.ts
AustinXII: Yes, the calc has status moves afterall

calc/src/mechanics/gen2.ts
AustinXII: Hm alright, the coding did the actual calculation looked weird fsr. Can you write tests though?

Enigami: "Write tests"? Not sure what you mean. I thought you meant showing that the results were correct, which I did (the Togepi and Togetic calcs came from tests from index.html after applying my changes), but you seem to want something else?

calc/src/mechanics/gen2.ts
Enigami: Ah, right. Fixed.

calc/src/data/moves.ts
Enigami: Alright, removed the delete.

calc/src/mechanics/gen2.ts
AustinXII: No we actually have test, see https://github.com/smogon/damage-calc/blob/master/calc/src/test/move.test.ts

Enigami: If I understand correctly, I needed to do a npm test? It gave me some issues when it got to the gts check at first, but I cleared them up and passes through the test fine. Or am I supposed to do something else?

AustinXII: No not that but yeah you should always `/npm test` before making a pr

AustinXII: See the commit I linked for the type of tests I’m talking about

Enigami:
I could use a little, I haven't made any before so I'm not exactly sure what I need to do. Is it supposed to look something like this?

    test('gen 2 present', () => {
      const attacker = new Pokemon(2, 'Togepi', {boosts: {level: 5, atk: -6}});
      const defender = new Pokemon(2, 'Umbreon', {boosts: {def: 6}});
      const move = new Move(2, 'Present');
      const result = calculate(2, attacker, defender, move);
      expect(result.damage).toBeRange(125, 147);
      expect(result.desc()).toBe(
        '-6 Togepi Present vs. +6 Umbreon: 125-147 (31.8 - 37.4%) -- 0% chance to 3HKO after Leftovers recovery'
      );
    });

I wanted to add a burn to Togepi, and maybe Reflect on Umbreon's side, since those should not have any effect on Present's damage output anyway, but I have not yet figured out how to do so.

scheibo:
```ts
new Pokemon(2, 'Togepi', {boosts: {level: 5, atk: -6}, status: 'Burned'});
...
const field = new Field({defenderSide: {isReflect: true}});
const result = calculate(2, attacker, defender, move, field);
```

should do the trick

Enigami:
Implemented a test for Present. I also noticed there was a mismatch between how Damage Calc calculated critical hit damage and how Showdown did. Damage Calc doubled damage, Showdown doubled level. After reviewing the Crystal disassembly, it appears it doubled damage and not level, so Showdown's critical hit damage calculation for Gen 2 is currently wrong.

From mods\gen2\scripts.js
// The move is a critical hit. Several things happen here.
		if (isCrit) {
			// Level is doubled for damage calculation.
			level *= 2; //<--- this is wrong

scheibo:
> After reviewing the Crystal disassembly, it appears it doubled damage and not level, so Showdown's critical hit damage calculation for Gen 2 is currently wrong.

I'm sure we'd welcome a PR there as well!

calc/src/desc.ts
```diff
@@ -524,7 +524,8 @@ function getEndOfTurn(
       damage -= gen > 6 ? Math.floor(defender.maxHP() / 32) : Math.floor(defender.maxHP() / 16);
       texts.push('reduced burn damage');
     } else if (!defender.hasAbility('Magic Guard')) {
-      damage -= gen > 6 ? Math.floor(defender.maxHP() / 16) : Math.floor(defender.maxHP() / 8);
+      damage -=
+        gen === 1 || gen > 6 ? Math.floor(defender.maxHP() / 16) : Math.floor(defender.maxHP() / 8);
```
scheibo:
```suggestion
         Math.floor(defender.maxHP() / (gen === 1 || gen > 6 ? 16 : 8));
```

Enigami: I noticed some of the other "damage -=" calculations also could be simplified by moving Math.floor(defender.maxHP() / {value}) out of the conditional operators, so I went ahead and changed those too

scheibo: Thank you! Thats awesome!

Enigami:
Oops! I forgot to run it through a test and accidentally missed a semi-colon at line 488 of calc/src/desc.ts. I fixed it on my end, but the pull request was merged before I did so.

PR #354, opened 2019-11-24
Add Gen 8 moves and Pokemon

KrisXV:


AustinXII:
Looks good. I’ll let you finish what ever you want to edit before merging

calc/src/move.ts
```diff
@@ -58,6 +63,56 @@ export class Move {
     let data: MoveData & {name: string} = extend(true, {name}, MOVES[gen][name], options.overrides);
 
     // If isZMove but there isn't a corresponding z-move, use the original move
+    if (options.useMax && 'maxPower' in data) {
+      const maxMoveName: string = getMaxMoveName(data.type, options.species, !!(data.category === 'Status'));
```
KrisXV: by the way i dont actually know where options.species or options.item would get assigned so this does almost nothing right now

calc/src/mechanics/gen8.ts
```diff
@@ -396,17 +396,17 @@ function calculateModern(
...
+      basePower = move.bp * (['Gmax', 'Dynamax'].indexOf(defender.name) !== -1 ? 2 : 1);
       description.moveBP = basePower;
       break;
     case 'Payback':
       basePower = turnOrder === 'LAST' ? 100 : 50;
       description.moveBP = basePower;
       break;
-    case 'Bolt Break':
+    case 'Bolt Beak':
     case 'Fishious Rend':
```
KrisXV: also this is throwing an error which is contributing to the failing build

`src/mechanics/gen8.ts:408:5 - error TS7029: Fallthrough case in switch.`

not sure why

KrisXV: im dumb

src/index.template.html
```diff
@@ -701,6 +701,17 @@
                                 <label class="btn btn-xwide" for="srR">Stealth Rock</label>
                             </div></td>
                         </tr>
+                        <tr class="gen-specific g4 g5 g6 g7 g8">
```
AustinXII: I don’t think steel spikes was around in gens 4, 5, 6, and 7 O.o

KrisXV: Done

PR #358, opened 2019-12-23
Fix Gen 8 using Gen 7's calculator

RoccoDev:
Due to `gen7.ts#calculateModern` and `gen8.ts#calculateModern` having the same name, Gen 7's function would still be used when `gen == 8`.

This also adds/fixes Punk Rock.


AustinXII:
Thanks!

PR #361, opened 2019-12-28
fix: Body Press

astapi:
atc, def are reversed. When using BodyPress

### DIFF
--- a/calc/src/mechanics/gen8.ts
+++ b/calc/src/mechanics/gen8.ts
@@ -702,7 +702,7 @@ function calculateGen8(
     move.category = attackSource.stats.atk > attackSource.stats.spa ? 'Physical' : 'Special';
   }
   const attackStat =
-    move.category === 'Special' ? 'spa' : move.name === 'Body Press' ? 'atk' : 'def';
+    move.category === 'Special' ? 'spa' : move.name === 'Body Press' ? 'def' : 'atk';
   description.attackEVs =
     attacker.evs[attackStat] +
     (NATURES[attacker.nature][0] === attackStat


AustinXII:
There’s something else going on if this wasn’t working before..

Zarel:
@AustinXII This seems correct to me. What's the problem?

AustinXII:
@Zarel this is how it was implemented at first when all the reports were first coming in. I changed it albeit due to a misread on my end but without “proof” this is actually fixing the issue, this shouldn’t be merged. The only thing I could think of is #358 fixed it

astapi:
@AustinXII 
In cases other than "Body Press", use "def".  
At least it seems correct to use "atk"

Why did you fix it here?
https://github.com/smogon/damage-calc/commit/b12e847ea5de861682854614883009577fbbafe9

AustinXII:
@astapi because of all the reports of it being broken. Can you screenshot this change working ?

calc/src/mechanics/gen8.ts
```diff
@@ -702,7 +702,7 @@ function calculateGen8(
     move.category = attackSource.stats.atk > attackSource.stats.spa ? 'Physical' : 'Special';
   }
   const attackStat =
-    move.category === 'Special' ? 'spa' : move.name === 'Body Press' ? 'atk' : 'def';
+    move.category === 'Special' ? 'spa' : move.name === 'Body Press' ? 'def' : 'atk';
```
Zarel: For readability reasons, we might as well do something like:

```suggestion
    move.name === 'Body Press' ? 'def' : (move.category === 'Special' ? 'spa' : 'atk');
```

Zarel:
@AustinXII I'm so confused right now, do you not understand your code in this commit and that reversion?

Zarel:
Do you understand that Body Press uses Def as the attacking stat, rather than Atk?

AustinXII:
@Zarel yes. This was how body press was originally done when we were first getting all the reports of body press being broken which is why I want to make sure this works now. Ignore my “fix” commit as I misread

Zarel:
Whether or not this commit is all that's needed to fix Body Press, it clearly moves in the right direction and fixes an obvious typo. I would merge it on those grounds alone.

AustinXII:
@Zarel but then we’re back to square one. My personal preference is to actually fix what’s broken instead of going back to what we had originally since that was broken as well

astapi:
@AustinXII 

The result of damage calculation when using Body Press. Looks good for me.
Can I find a broken report somewhere?

## case 1

Attacker: Corviknight
Defender: Corviknight
move: Body Press

Attacker Corviknight stats

```
hp: 173
atk: 107
def: 125
spa: 73
spd: 105
spe: 87
```

The "def" of 125 is used as the attack.

## case 2

Attacker: Corviknight { boosts: { def:2 } }
Defender: Corviknight
move: Body Press

Attacker Corviknight stats

```
hp: 173
atk: 107
def: 250
spa: 73
spd: 105
spe: 87
```

The "def" of 250 is used as the attack.

## code

```
import * as calc from './calc/index'

let attacker = new calc.Pokemon(8, 'Corviknight', { 
  level: 50,
});
const defender = new calc.Pokemon(8, 'Corviknight', { level: 50 });
const move1 = new calc.Move(8, 'Body Press', { ability: attacker.ability });
const calcResult = calc.calculate(8, attacker, defender, move1);
console.log('Result1');
console.log(calcResult.desc());

attacker = new calc.Pokemon(8, 'Corviknight', { 
  level: 50,
  boosts: { def: 2, }
});
const calcResult2 = calc.calculate(8, attacker, defender, move1);
console.log('Result2');
console.log(calcResult2.desc());
```

## Result

```
Result1
0 Def Corviknight Body Press vs. 0 HP / 0 Def Corviknight: 31-37 (17.9 - 21.3%) -- possible 5HKO
Result2
+2 0 Def Corviknight Body Press vs. 0 HP / 0 Def Corviknight: 61-72 (35.2 - 41.6%) -- guaranteed 3HKO
```

AustinXII:
Hmm looks good to me, if people start complaining about this not working again then idk

astapi:
@AustinXII  thanks !! 😆

kyleoliveiro:
Has this change been deployed to production? As of now it is still using the Atk stat.

Zarel:
I pulled and ran the build script. @scheibo is there anything else I need to do?

scheibo:
`cd /var/www/html/calc.pokemonshowdown.com/ && git sync && npm run compile`. Im assuming you built it wrong.

PR #362, opened 2019-12-31
Add Dauntless Shield + Fixes

tmagicturtle:
Right now, Intrepid Sword and Gorilla Tactics boost special attacks on the calc as well. This change fixes that, and also introduces Dauntless Shield to the damage calculation.

### DIFF
--- a/calc/src/mechanics/gen8.ts
+++ b/calc/src/mechanics/gen8.ts
@@ -783,7 +783,7 @@ function calculateGen8(
   ) {
     atMods.push(0x2000);
     description.attackerAbility = attacker.ability;
-  } else if (attacker.hasAbility('Gorilla Tactics', 'Intrepid Sword')) {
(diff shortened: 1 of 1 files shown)


calc/src/mechanics/gen8.ts
```diff
@@ -783,7 +783,10 @@ function calculateGen8(
   ) {
     atMods.push(0x2000);
     description.attackerAbility = attacker.ability;
-  } else if (attacker.hasAbility('Gorilla Tactics', 'Intrepid Sword')) {
+  } else if (
```
AustinXII: I don’t like this formatting tbh :(

AustinXII: I swear we have longer lines than this in this file

AustinXII:
super seeded by #363 and #30025221489bb8d224ff4e343e9c25943166738f

PR #363, opened 2020-01-07
Fix Intrepid Sword

asgdf:
It is not a 1.5x damage boost, it just gives +1 on switchin.
(I assumed there was a reason all these boost-generating ability checks are separate functions, so I made a separate one as well)


asgdf:
Ok I ended up doing the multiline split. Gts is happy now.

AustinXII:
Yay!

PR #364, opened 2020-01-07
Import Dynamax Support

AustinXII:


calc/src/data/moves.ts
```diff
@@ -35,7 +35,9 @@ export interface MoveData {
   readonly isTwoHit?: boolean;
 }
 
-const RBY: {[name: string]: MoveData} = {
+const RBY: {
```
AustinXII: @scheibo although it doesn't overly matter, are you ok with this styling ?

AustinXII: if so this should be good to mergee

scheibo: It's enforced automatically. I'm not in love with the fact that Prettier chooses to style things this way, but its not something we can configure anyway.

PR #368, opened 2020-01-23
fix multi hit ko-chance

AustinXII:


AustinXII:
More hacky version of #369

PR #371, opened 2020-02-11
Fix Gorilla Tactics

RoccoDev:
The issue was in the second branch of control flow:
```js
['Gmax', 'Dynamax'].indexOf(defender.name) !== -1
```
which doesn't make sense since Gorilla Tactics doesn't work only if the **attacker** is **not** dynamaxed.

### DIFF
--- a/calc/src/mechanics/gen8.ts
+++ b/calc/src/mechanics/gen8.ts
@@ -791,7 +791,7 @@ function calculateGen8(
     description.attackerAbility = attacker.ability;
   } else if (
     attacker.hasAbility('Gorilla Tactics') &&
-    ['Gmax', 'Dynamax'].indexOf(defender.name) !== -1
+    ['Gmax', 'Dynamax'].indexOf(attacker.name) === -1
   ) {
     atMods.push(0x1800);
     description.attackerAbility = attacker.ability;
--- a/calc/src/mechanics/gen8.ts
+++ b/calc/src/mechanics/gen8.ts
@@ -35,12 +35,12 @@ const SS = 8;
 
 export function makeCalculate(gen: 5 | 6 | 7 | 8) {
   return (attacker: Pokemon, defender: Pokemon, move: Move, field: Field) =>
-    calculateModern(gen, attacker, defender, move, field);
+    calculateGen8(gen, attacker, defender, move, field);
 }
 
 export const calculateSS = makeCalculate(SS);
 
-function calculateModern(
+function calculateGen8(
   gen: 5 | 6 | 7 | 8,
   attacker: Pokemon,
   defender: Pokemon,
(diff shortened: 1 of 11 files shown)


calc/src/mechanics/gen8.ts
```diff
@@ -791,7 +788,7 @@ function calculateGen8(
     description.attackerAbility = attacker.ability;
   } else if (
     attacker.hasAbility('Gorilla Tactics') &&
-    ['Gmax', 'Dynamax'].indexOf(defender.name) !== -1
+    ['Gmax', 'Dynamax'].indexOf(attacker.name) === -1
```
Zarel: Is the attacker name literally going to be "Gmax" or "Dynamax"? I think this clearly needs two passes, so they can be searched.

RoccoDev: I agree that we shouldn't use names to track maxing and should instead use a field to track it. What's the state of the Dynamax branch?

Zarel: I have no clue, sorry, in this repo I'm just a rando who codes.

calc/src/mechanics/gen8.ts
```diff
@@ -1053,6 +1050,10 @@ function calculateGen8(
     finalMods.push(0x800);
     description.defenderAbility = defender.ability;
   }
+  if (move.isSound && attacker.hasAbility('Punk Rock')) {
+    finalMods.push(0x1300);
```
AustinXII: What’s wrong with 0x14cd?

AustinXII: Also what do you mean by “checked twice”?

RoccoDev: You're right, my bad, that was probably some leftover from testing. Changed it back to 0x14cd.

AustinXII:
The only thing left is finding out what you meant by the  gorillla tactics check being checked twice

AustinXII:
(I’m going off of one of your commits name btw)

RoccoDev:
Yeah so, when I merged, I accidentally left another check for Punk Rock (the one with 0x1300) in the code. So when it was implemented elsewhere, both checks were present (in my local repository).
That commit in tandem with the latest one (that fixes 0x14cd) simply do nothing.

AustinXII:
Oooh makes sense. Thanks!

calc/src/mechanics/gen8.ts
AustinXII: I can add a field to dynamax pokemon and update this once that branch is done

RoccoDev:
I made something similar in #372

calc/src/mechanics/gen8.ts
RoccoDev: I made something similar in #372

calc/src/mechanics/gen8.ts
AustinXII: Sweet, just check the field for this check then :)

PR #374, opened 2020-03-06
Remove Side#isDynamaxed and rename Pokemon#isMax

scheibo:
`Side#isDynamaxed` seems completely wrong (its a field in the Pokemon controls, not field) and doesn't belong there in the calc. But the field has a more descriptive name, so I've switched to that.


scheibo:
K so the last two commits actually fix this. I can maybe see what the original author just threw this on Side - Side is immutable so doesnt get cloned

PR #376, opened 2020-03-07
Use isMax option in constructor

ajhyndman:
It looks like `isMax` is accepted as an option by the `Pokemon` class constructor but not actually used.

It looks like this might be an oversight!  This change sets the current value of this.isMax using that option.


scheibo:
Thanks for catching this! https://github.com/smogon/damage-calc/pull/374 is already open with changes in this area so I'm going to include your change there.

ajhyndman:
Amazing!  Thanks for rolling that in.

Issue #348, opened 2019-09-04
Miscellaneous Bugs & Feature Requests


scheibo:
Past reported bugs and feature requests from the [Smogon thread](https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/). I'm unsure if the 'Potential' bugs are still relevant, they should be moved under 'Verified' once confirmed (or removed).

Bugs can still be addressed as they are reported on the thread and is still the correct venue for reporting bugs, but this issue just exists to make track things which may have fallen through the cracks and been forgotten about.

# Bugs

## Verified

- [ ] **Parental Bond**: https://github.com/Zarel/honko-damagecalc/issues/330
- [ ] **Gen 7 modifer order**: https://github.com/Zarel/honko-damagecalc/issues/331

## Potential

- [ ] **Aerilate in "All vs One"**: _"Noticed that Aerilate doesn't seem to be
  calculating correctly on "All vs One". Mega Blaziken getting hit with Mega
  Salamence's Double-Edge only does 73% damage apparently..."_
- [ ] **Recoil damage**:  _"There seems to be a problem with recoil calculation when the damage done by a recoil move exceeds the targets maximum HP, causing it to become something close to 0. Please either fix or implement this on PS."_
- [ ] **Normalize on moves already Normal type**:  _"Normalize doesn't add the 20% bonus damage on moves that are already normal type"_
- [ ] **Stamina does not work with Multi-Hits moves**: _"I know only mudsdale
  has it, but as a PU player, it would help me a lot since I always have to calc
  in my head how low will mudsdale be after two hits, the second hit being
  physical: first calc the first hit, then put +1 in defense and calc a second
  hit."_
- [ ] **KO chance broken for repeated use of moves which drop stats**
    - _"there seems to be a bug in how the calculator works out the % chance to kill of repeated use of moves that weaken offensive stats (like Superpower and Draco Meteor for example). It seems to show the reverse percentage of what it should, as in, say if a move has a 75% chance to 2HKO, it will say it has a 25% chance to 2HKO."_
    - _"When Arcanine uses Overheat multiple times, the calc inverts the % chance to KO with respect to additional EVs."_
[...]

PR #377, opened 2020-03-07
Populate neutral nature stat associations

ajhyndman:
Technically, the neutral natures are each associated with a plus and minus stat of the same value.  This would be nice to include, for lookup purposes.

I'm not 100% sure if this will break anything though, so definitely need another pair of eyes.

Associations sourced from: https://pokemondb.net/mechanics/natures

### DIFF
--- a/calc/src/data/natures.ts
+++ b/calc/src/data/natures.ts
@@ -3,14 +3,14 @@ import {Stat} from '../stats';
 
 export const NATURES: {[name: string]: [Stat?, Stat?]} = {
   Adamant: ['atk', 'spa'],
-  Bashful: [undefined, undefined],
+  Bashful: ['spa', 'spa'],
   Bold: ['def', 'atk'],
   Brave: ['atk', 'spe'],
   Calm: ['spd', 'atk'],
   Careful: ['spd', 'spa'],
-  Docile: [undefined, undefined],
+  Docile: ['def', 'def'],
   Gentle: ['spd', 'def'],
-  Hardy: [undefined, undefined],
+  Hardy: ['atk', 'atk'],
   Hasty: ['spe', 'def'],
   Impish: ['def', 'spa'],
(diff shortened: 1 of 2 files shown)


scheibo:
This will break `calcStat`, as the failing tests indicate. I'm also not sure this provides any value because its not displayed to the UI and because to make the implementation work with the plus and minus both set the same youd need to do more work than just checking for the absence.

We plan to [switch to a different implementation of `Nature`](https://github.com/smogon/data/blob/master/src/natures.ts), but it doesnt represent both the plus and minus of the same stat either.

I appreciate your interest in contributing - have you found https://github.com/smogon/damage-calc/issues/348 with ideas for where you might help out? Please also feel free to join http://spo.ink/dev and ask in #other-dev if you want to get involved.

ajhyndman:
Thanks for the context! 

I just discovered that this library was published to npm, and was inspired to try my hand at my own calculator interface: https://github.com/ajhyndman/visual-pokemon-calc

In my app, I'm trying out allowing nature selection from the stat, instead of as a single list of natures.  To do that, I wanted to support reverse lookup on all of the natures.

https://github.com/ajhyndman/visual-pokemon-calc/blob/master/src/PokemonPicker.tsx#L104-L109

I can easily patch the NATURES object for my purposes though, so if it makes things messier here, it's no big deal.

ajhyndman:
I pushed a fix for calcStat, but I won't take it personally if you prefer to close this!

scheibo:
`prettier`s formatting of this is regrettable, but I'm probably going to switch out the formatter anyway so I won't make you wrestle with it.

scheibo:
This seems like it actually causes some issues with how the descriptions are displayed:

https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/post-8401679

PR #381, opened 2020-03-24
Abstract out the data layer and make it adaptable

scheibo:
With this change, users will be able to opt out of using `@smogon/calc`'s data files by instead requiring`@smogon/calc/adaptable` and implementing their own implementation of `Generation` that is implements the `@smogon/calc/data/interface`. There are at least 3-4 different data layers floating around (`@smogon/calc/data`, `smogon/pokemon-showdown`, `smogon/pokemon-showdown-client`, `@smogon/data`), and making the calc's data layer optional allows it to be used in different contexts without requiring duplicate data files to be downloaded (ie. `smogon/calc/data` can be easily stripped). Furthermore, these other data layers are potentially better maintained and could be swapped in to improve accuracy.

In the future, smogon/damage-calc would like to tear out its own data layer and instead depend on `@smogon/data`(and ideally in such a future `smogon/pokemon-showdown` and `smogon/pokemon-showdown-client` would also align on this one true data layer and thus no adapter layer is actually required). This change is more incremental - this simplifies switching to `@smogon/data` at some point but does not commit full bore.

Several **breaking changes** were made in the process of this reorganized (and likely some regressions were introduced), primarily because it simplified things. This package is still 0.X, so breaking changes are par for the course. 

 - `mechanics/gen{5,6,7,8}.ts` were all deleted, with `mechanics/modern.ts` replacing all of them. Originally the naming scheme was for consistency, but clear this resulted in confusion as the `gen7.ts` file was simply forked into `gen8.ts` instead of being made to delegate to `gen8.ts` as was designed. Hopefully now things are more clear (plus less to download)
  - there will be a slight performance hit from `new`-ing up a bunch of objects when first loading `smogon/calc/data`, this isn't really expected to be noticeable and is a startup cost (and in the official client will be a wash with not downloading `gen7.ts`)
  - the type name `None` has been more correctly renamed `???`
  - the gender enum has changed to `'M' | 'F' | 'N'` like PS.
  - the non-bundled loading order of the compiled files has changed (and is still hacky and fragile = use a bundler!). Its possibly something broke here - I'm pretty sure the view code was placed where it was previously because of a function name conflict, but I couldn't repro...
  - `Generation` is now an object and not a number (used `Generation#num` for the number). This is obviously the largest consequence, as `Generation` is also the data entry point. Care has been taken to ensure the simpler APIs just using a number have been preserved, partly for backwards compatibility but also simply for convenience in testing.
  - a `Stats` singleton is now exported instead of several loose functions

Most field name changes have been avoided in this change, but future changes to make the data layer more aligned with `@smogon/data` / PS will further break API compatibility (eg. renaming the fields in `SpeciesData`)


scheibo:
Courtesy heads up to @ajhyndman :)

scheibo:
https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/post-8402280 points to maybe `gen8.ts` (and now `modern.ts` not having zmove support?

scheibo:
I'm doing some more changes in the calc that will cause conflicts with this so I'm merging @AustinXII, if you have comments I will address them in a followup

PR #383, opened 2020-03-29
Tweaking move data for Rolling Kick

jawgin:
Rolling kick is herein changed to be 60 base power.

Was originally listed as 70 base power, but every source on the web (and indeed, its data entry in the Pokemon Showdown files) points to it being 60 instead.


AustinXII:
thanks!

PR #393, opened 2020-05-10
Fix Max Moves to inflict reduced damage through Protect

yghdj:
Using a Max Move on a protected target didn't show the right amount of damage, acting as if the max move was a regular move (i.e. 0 damage). Now it should be displaying the damage of the move divided by 4. I just used what was already coded for Z moves.

### DIFF
--- a/calc/src/mechanics/gen78.ts
+++ b/calc/src/mechanics/gen78.ts
@@ -87,7 +87,7 @@ export function calculateSMSS(
     return result;
   }
 
-  if (field.defenderSide.isProtected && !move.breaksProtect && !move.isZ) {
+  if (field.defenderSide.isProtected && !move.breaksProtect && !move.isZ && !attacker.isDynamaxed) {
     desc.isProtected = true;
     return result;
   }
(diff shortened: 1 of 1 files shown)


calc/src/mechanics/gen78.ts
```diff
@@ -1008,6 +1008,11 @@ export function calculateSMSS(
     desc.isProtected = true;
   }
 
+  if (field.defenderSide.isProtected && attacker.isDynamaxed) {
```
scheibo, line 1011: I think we should actually be able to combine this conditional with the `if` on line 1006? Something like

```ts
if (field.defenderSide.isProtected &&
	((move.isZ && attacker.item && attacker.item.includes(' Z')) ||
	  attacker.isDynamaxed)) {
```

yghdj: We can do that, i just thought it would be more readable this way?

AustinXII: Considering the lines are doing the same thing, I don’t think readability is a concern. We also have much longer if statements than this so we should be good :)

yghdj: Alright, just applied the change.

scheibo:
Thanks, @yghdj!

PR #396, opened 2020-05-31
Fix display issue when using GMax formes

yghdj:
I noticed while working on adding specific EOT effects from GMax moves that those would not appear if dynamaxing a GMax forme. For example, selecting Charizard-Gmax standard set, then checking the Dynamax button, would transform Flamethrower into Max Flare instead of G-Max Wildfire.

The effect was purely visual before, so it wasn't much of an issue, but since EOT effects such as G-Max Wildfire will be added to the calculator in the near future, it would be interesting to have this behaving correctly now. This should hopefully be done now.

### DIFF
--- a/calc/src/move.ts
+++ b/calc/src/move.ts
@@ -99,6 +99,7 @@ export class Move implements State.Move {
     this.useZ = options.useZ;
     this.useMax = options.useMax;
     this.overrides = options.overrides;
+    this.species = options.species;
 
     this.bp = data.basePower;
     // These moves have a type type of these moves exists, but the damage they deal is typeless so we override it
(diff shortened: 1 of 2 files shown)


calc/src/move.ts
```diff
@@ -223,6 +224,7 @@ const ZMOVES_TYPING: {
 };
 
 export function getMaxMoveName(moveType: I.TypeName, pokemonSpecies?: string, isStatus?: boolean) {
+  console.log('getMaxMoveName');
```
scheibo, line 227: Please remove :)

yghdj:
Yup, it's removed!

scheibo:
Thanks! I'll take a look at your other PR in a soon, this one was just shorter and I'm eating lunch :)

PR #399, opened 2020-06-21
Give new Gmax forms accurate abilities

ajhyndman:
It looks like a few copy-pasted abilities snuck in.

### DIFF
--- a/calc/src/data/species.ts
+++ b/calc/src/data/species.ts
@@ -8236,7 +8236,7 @@ const SS_PATCH: {[name: string]: DeepPartial<SpeciesData>} = {
   'Rillaboom-Gmax': {
     types: ['Grass'],
     bs: {hp: 100, at: 125, df: 90, sa: 60, sd: 70, sp: 85},
-    weightkg: 80,
+    weightkg: 0,
     abilities: {0: 'Overgrow'},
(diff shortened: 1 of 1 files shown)


calc/src/data/species.ts
```diff
@@ -8454,7 +8454,7 @@ const SS_PATCH: {[name: string]: DeepPartial<SpeciesData>} = {
     types: ['Poison', 'Fairy'],
```
KrisXV, line 8454:
```suggestion
    types: ['Grass', 'Poison'],
```

calc/src/data/species.ts
```diff
@@ -8237,7 +8237,7 @@ const SS_PATCH: {[name: string]: DeepPartial<SpeciesData>} = {
     types: ['Grass'],
     bs: {hp: 100, at: 125, df: 90, sa: 60, sd: 70, sp: 85},
     weightkg: 80,
```
KrisXV, line 8239:
```suggestion
    weightkg: 0,
```

ajhyndman: Looks like urshifu slipped in as well.  Good catch.

scheibo:
Thanks, this was fixed by 397b263

PR #400, opened 2020-06-28
Tutor move damage updates

ajhyndman:
## Expanding Force

Serebii and Bulbapedia disagree on the damage for expanding force.  According to Serebii, the base power is boosted by 1.5x in Psychic Terrain.  According to Bulbapedia, the terrain modifier is _replaced_ by a 2x modifier.

However, a base power increase of 1.5 results in 1.5 * 1.3 = 1.95 total damage modifier — which is very close to 2x.  Until we have clear data on exactly which is correct, I opted for 1.5 damage modifier under terrain.  This would be consistent with other moves in the Isle of Armor release, and is much closer to accurate than the calculator is returning now.

https://bulbapedia.bulbagarden.net/wiki/Expanding_Force_(move)
https://www.serebii.net/attackdex-swsh/expandingforce.shtml

## Misty Explosion

Serebii and Bulbapedia both now agree that Misty Explosion is boosted by 1.5x under terrain.

https://bulbapedia.bulbagarden.net/wiki/Misty_Explosion_(move)
https://www.serebii.net/attackdex-swsh/mistyexplosion.shtml

### DIFF
--- a/calc/src/mechanics/gen78.ts
+++ b/calc/src/mechanics/gen78.ts
@@ -1,4 +1,4 @@
-﻿import {Generation, AbilityName, StatName, Terrain} from '../data/interface';
+import {Generation, AbilityName, StatName, Terrain} from '../data/interface';
 import {toID} from '../util';
 import {
   getBerryResistType,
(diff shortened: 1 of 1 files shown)


KrisXV:
The Expanding Force mechanics are not confirmed, and serebii and bulbapedia are not reliable sources for nitty-gritty mechanics.

ajhyndman:
> The 1.5x boost on top of the 1.3x terrain boost is what should be implemented for now.

Great, that's what I've done in this pull request.

AustinXII:
Sure, sounds good

PR #401, opened 2020-06-28
Fix Triple Axel's base power boost

KrisXV:


scheibo:
Thanks

AustinXII:
Is this right? If the move hits twice the damage is going to be 40+40 and for 3 hits it’s going to be 60+60+60? I did 30 because 2 hits should be 40+20(30x2) and for three hits 60+40+30(40x3)

AustinXII:
Although I am freshly out of surgery and on medications so forgive me if am wrong lol

AustinXII:
You’re modifying the base power of each hit not the final power

KrisXV:
Oh... I didn’t realize just HOW hacky this was

AustinXII:
Should’ve went through with my more detailed comment to explain why it’s listed like this. Wasn’t expecting a sneaky revive from pre lol

PR #403, opened 2020-06-29
Fix some move mechanics

KrisXV:


AustinXII:
Sweet, just curious how’s the UI looking with this change? The middle portion must be getting pretty big

KrisXV:
@AustinXII that's why i split it up into two sections

img

PR #404, opened 2020-07-27
Fix Shell Side Arm's category

ajhyndman:
Based on the best science I can find, Shell Side Arm's damage category is dependant on the relative _predicted damage_ and not just the defense stats of the target pokemon.

This change matches that behaviour.  I left some notes in my implementation in case this needs following up on, once there is better info.

### DIFF
--- a/calc/.eslintrc
+++ b/calc/.eslintrc
@@ -2,7 +2,6 @@
   "root": true,
   "extends": "@pkmn",
   "rules": {
-    "max-len": ["error", { "code": 100, "ignoreUrls": true }],
     "no-shadow": 0,
     "jest/no-standalone-expect": "off",
     "@typescript-eslint/restrict-template-expressions": "off"
--- a/calc/.eslintrc
+++ b/calc/.eslintrc
@@ -2,6 +2,7 @@
   "root": true,
   "extends": "@pkmn",
   "rules": {
+    "max-len": ["error", { "code": 100, "ignoreUrls": true }],
     "no-shadow": 0,
     "jest/no-standalone-expect": "off",
     "@typescript-eslint/restrict-template-expressions": "off"
--- a/calc/src/mechanics/gen78.ts
+++ b/calc/src/mechanics/gen78.ts
@@ -717,14 +717,14 @@ export function calculateSMSS(
     move.category = attackSource.stats.atk > attackSource.stats.spa ? 'Physical' : 'Special';
   }
   const attackStat =
-    move.named("Shell Side Arm") &&
-    getShellSideArmCategory(attacker, defender) === "Physical"
-      ? "atk"
-      : move.named("Body Press")
-      ? "def"
-      : move.category === "Special"
-      ? "spa"
-      : "atk";
+    move.named('Shell Side Arm') &&
+    getShellSideArmCategory(attacker, defender) === 'Physical'
+      ? 'atk'
+      : move.named('Body Press')
+        ? 'def'
+        : move.category === 'Special'
+          ? 'spa'
+          : 'atk';
   desc.attackEVs =
     move.named('Foul Play')
       ? getEVDescriptionText(gen, defender, attackStat, defender.nature)
--- a/calc/src/mechanics/gen78.ts
+++ b/calc/src/mechanics/gen78.ts
@@ -33,6 +33,7 @@ import {
   getFinalDamage,
   getModifiedStat,
   getMoveEffectiveness,
+  getShellSideArmCategory,
   getWeightFactor,
   handleFixedDamageMoves,
   isGrounded,
@@ -716,9 +717,14 @@ export function calculateSMSS(
     move.category = attackSource.stats.atk > attackSource.stats.spa ? 'Physical' : 'Special';
   }
   const attackStat =
-    (move.named('Shell Side Arm') && defender.stats.def < defender.stats.spd) ? 'atk'
-    : move.named('Body Press') ? 'def'
-    : move.category === 'Special' ? 'spa' : 'atk';
+    move.named("Shell Side Arm") &&
+    getShellSideArmCategory(attacker, defender) === "Physical"
+      ? "atk"
+      : move.named("Body Press")
+      ? "def"
+      : move.category === "Special"
+      ? "spa"
+      : "atk";
   desc.attackEVs =
     move.named('Foul Play')
       ? getEVDescriptionText(gen, defender, attackStat, defender.nature)
@@ -835,7 +841,7 @@ export function calculateSMSS(
 
   let defense: number;
   const hitsPhysical = move.defensiveCategory === 'Physical' ||
-    (move.named('Shell Side Arm') && defender.stats.def < defender.stats.spd);
+    (move.named('Shell Side Arm') && getShellSideArmCategory(attacker, defender) === 'Physical');
   const defenseStat = hitsPhysical ? 'def' : 'spd';
   desc.defenseEVs = getEVDescriptionText(gen, defender, defenseStat, defender.nature);
   if (defender.boosts[defenseStat] === 0 ||
--- a/calc/src/mechanics/util.ts
+++ b/calc/src/mechanics/util.ts
@@ -358,9 +358,7 @@ export function getFinalDamage(
 }
 
 /**
- * Determines which move category Shell Side Arm should behave as. Applications
- * that consume this library may want to import this function to show users move
- * category information.
+ * Determines which move category Shell Side Arm should behave as.
  *
  * Notes:
  * 1. As far as I can tell, research hasn't been tone on whether it's the larger
@@ -369,11 +367,13 @@ export function getFinalDamage(
  * Ratio seems more likely to me, since the pokemon damage formula is
  * dependent on the ratio and not the difference of attack and defense stats.
  *
- * 2. in game, when stats are equal, category is random.  For damage, it won't
+ * 2. In game, when stats are equal, category is random.  For damage, it won't
  * matter most of the time (items like assault vest aren't included in this
- * decision though, so *may* take effect).
+ * decision though, so *may* take effect). This implementation defaults to
+ * 'Special'.
  *
- * {@link https://www.reddit.com/r/VGC/comments/hcfe4k/galarian_slowbro_shell_side_arm_breakdown/}
+ * See also:
+ * {@link https://github.com/smogon/pokemon-showdown/commit/65d2bb5d0c36b61a2577382cc30aa861521113fd#diff-706ad2c9c5d1920e4245d5debe9e3c11}
  *
  * @param source Attacking pokemon (after stat modifications)
  * @param target Target pokemon (after stat modifications)
--- a/calc/src/mechanics/util.ts
+++ b/calc/src/mechanics/util.ts
@@ -360,7 +360,7 @@ export function getFinalDamage(
 /**
  * Determines which move category Shell Side Arm should behave as.
  *
- * A simplified formula can be used here compared to what the research 
+ * A simplified formula can be used here compared to what the research
  * suggests as we do not want to implement the random tiebreak element of
  * move - instead we simply default to 'Special' and allow the user to override
  * this by manually adjusting the move's category.
--- a/calc/src/mechanics/util.ts
+++ b/calc/src/mechanics/util.ts
@@ -373,7 +373,7 @@ export function getFinalDamage(
  * 'Special'.
  *
  * See also:
- * {@link https://github.com/smogon/pokemon-showdown/commit/65d2bb5d0c36b61a2577382cc30aa861521113fd#diff-706ad2c9c5d1920e4245d5debe9e3c11}
+ * {@link https://github.com/smogon/pokemon-showdown/commit/65d2bb5d}
  *
  * @param source Attacking pokemon (after stat modifications)
  * @param target Target pokemon (after stat modifications)
--- a/calc/src/mechanics/util.ts
+++ b/calc/src/mechanics/util.ts
@@ -2,6 +2,7 @@ import {
   Generation,
   ID,
   ItemName,
+  MoveCategory,
   NatureName,
   StatName,
   StatsTable,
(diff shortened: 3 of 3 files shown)


calc/src/mechanics/util.ts
```diff
@@ -356,6 +357,34 @@ export function getFinalDamage(
...
+ * {@link https://www.reddit.com/r/VGC/comments/hcfe4k/galarian_slowbro_shell_side_arm_breakdown/}
+ *
+ * @param source Attacking pokemon (after stat modifications)
+ * @param target Target pokemon (after stat modifications)
+ * @returns 'Physical' | 'Special'
+ */
+export function getShellSideArmCategory(source: Pokemon, target: Pokemon): MoveCategory  {
+  const physicalDamage = source.stats.atk / target.stats.def;
+  const specialDamage = source.stats.spa / target.stats.spd;
+  return physicalDamage > specialDamage ? 'Physical' : 'Special'
```
KrisXV, line 385: This is not the correct formula.
```suggestion
  const physicalDamage = ((((2 * source.level / 5 + 2) * 90 * source.stats.atk) / target.stats.def) / 50);
  const specialDamage = ((((2 * source.level / 5 + 2) * 90 * source.stats.spa) / target.stats.spd) / 50);
  return physicalDamage > specialDamage ||
    (physicalDamage === specialDamage && Math.random() < 0.5)
    ? 'Physical'
    : 'Special';
```

(I'm unsure if this is the right way to do a random choice to simulate a coinflip)

calc/src/mechanics/util.ts
```diff
@@ -356,6 +357,34 @@ export function getFinalDamage(
   return OF16(pokeRound(Math.max(1, OF32(damageAmount * finalMod) / 0x1000)));
 }
 
+/**
+ * Determines which move category Shell Side Arm should behave as. Applications
+ * that consume this library may want to import this function to show users move
```
scheibo, line 362: https://github.com/smogon/damage-calc/blob/master/calc/src/adaptable.ts and https://github.com/smogon/damage-calc/blob/master/calc/src/index.ts is the external API for this package, any other `export`'s are internal. This method is not really of appropriate scope to be included in the top level external API. While you can still depend on internal implementation details, because its not part of the external API there are not going to be guarantees on its existence/behavior.

calc/src/mechanics/util.ts
```diff
@@ -356,6 +357,34 @@ export function getFinalDamage(
...
+ * category information.
+ *
+ * Notes:
```
scheibo, line 365: @Marty-D / @DaWoblefet

calc/src/mechanics/util.ts
scheibo: Damage cannot be stochastic. You have two choices here:

- either simply favor one over the other and leave a comment explaining this decision = the approach ajhyndman originally went with
- add a field to `Move` to configure how the tie should be broken (so that it could be wired up to a UI which allows the user to decide) = more correct, but potentially overkill

calc/src/mechanics/gen78.ts
```diff
@@ -716,9 +717,14 @@ export function calculateSMSS(
     move.category = attackSource.stats.atk > attackSource.stats.spa ? 'Physical' : 'Special';
   }
   const attackStat =
-    (move.named('Shell Side Arm') && defender.stats.def < defender.stats.spd) ? 'atk'
-    : move.named('Body Press') ? 'def'
-    : move.category === 'Special' ? 'spa' : 'atk';
+    move.named("Shell Side Arm") &&
```
scheibo, line 720: You should be able to just run `npm run fix` to handle these :)

calc/src/mechanics/util.ts
```diff
@@ -356,6 +357,34 @@ export function getFinalDamage(
...
+ * ratio (atk / def) or the difference (atk - def) which decides the move
+ * category.
+ * Ratio seems more likely to me, since the pokemon damage formula is
+ * dependent on the ratio and not the difference of attack and defense stats.
+ *
+ * 2. in game, when stats are equal, category is random.  For damage, it won't
+ * matter most of the time (items like assault vest aren't included in this
+ * decision though, so *may* take effect).
+ *
+ * {@link https://www.reddit.com/r/VGC/comments/hcfe4k/galarian_slowbro_shell_side_arm_breakdown/}
```
KrisXV, line 376: https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/page-51#post-8514607

Here is a post containing the research on the move.

calc/src/mechanics/util.ts
scheibo: Kris pointed me at https://github.com/smogon/pokemon-showdown/commit/65d2bb5d0c36b61a2577382cc30aa861521113fd#diff-706ad2c9c5d1920e4245d5debe9e3c11 commit. I believe the follow are the links to the research:

- https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/page-50#post-8509843
- https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/page-51#post-8511267
- https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/page-51#post-8514316
- https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/page-51#post-8514607

(man, how awesome are our researchers??)

calc/src/mechanics/gen78.ts
ajhyndman: Thanks!  I wasn't sure how to format.

calc/src/mechanics/util.ts
ajhyndman: Thanks for digging these up for me!

I _do_ think it's a pretty safe assumption that the comparison is based on real predicted damage.  But I can't help noticing that it still doesn't look like anyone has tested Shell Side Arm with an attacker/defender with stats something like:

Attacker:
Atk 100
SpA 50

Defender:
Def 50
SpD 25

I can't think of any other way to rule out that the move choice is based on a flat _difference_ of attack stat minus defense stat.

calc/src/mechanics/util.ts
ajhyndman: >  const physicalDamage = ((((2 * source.level / 5 + 2) * 90 * source.stats.atk) / target.stats.def) / 50);
>  const specialDamage = ((((2 * source.level / 5 + 2) * 90 * source.stats.spa) / target.stats.spd) / 50);

I recognise that this is technically correct, but since every other value is the same in both formulas, the relative ratio of attack / def is sufficient to determine which will be more damage.

To do the long form:

```
// physical
= ((((2 * source.level / 5 + 2) * 90 * source.stats.atk) / target.stats.def) / 50);
= ((((2 * 50 / 5 + 2) * 90 * source.stats.atk) / target.stats.def) / 50);
= ((((100 / 5 + 2) * 90 * source.stats.atk) / target.stats.def) / 50);
= ((((20 + 2) * 90 * source.stats.atk) / target.stats.def) / 50);
= (((22 * 90 * source.stats.atk) / target.stats.def) / 50);
= (((22 * 90 * source.stats.atk) / target.stats.def) / 50);
= (((1980 * source.stats.atk) / target.stats.def) / 50);
= ((1980 * (source.stats.atk / target.stats.def) / 50)
= ((1980 / 50) * (source.stats.atk / target.stats.def))
= (39.6 * (source.stats.atk / target.stats.def))

// special
= ((((2 * source.level / 5 + 2) * 90 * source.stats.spa) / target.stats.spd) / 50);
// ...
= (39.6 * (source.stats.spa / target.stats.spd))

// now do the comparison
(39.6 * (source.stats.atk / target.stats.def)) > (39.6 * (source.stats.spa / target.stats.spd))
// divide both sides by a postive real number (no need to reverse the direction of the inequality)
(source.stats.atk / target.stats.def)) > (source.stats.spa / target.stats.spd)
```

I noticed that the showdown implementation does the discrete Math.floor at each stage of computation, which I suppose _could_ matter for some close calls.  But I didn't see any research on those specific cases to verify that is in fact important.

calc/src/mechanics/util.ts
ajhyndman: It's definitely going to be useful to me, but I accept this reasoning.  I'll remove that line from the comment.

calc/.eslintrc
```diff
@@ -2,8 +2,9 @@
   "root": true,
   "extends": "@pkmn",
   "rules": {
+    "max-len": ["error", { "code": 100, "ignoreUrls": true }],
```
ajhyndman, line 5: Is this rule change worth upstreaming?

calc/src/mechanics/util.ts
ajhyndman: I don't expect anyone cares about this much detail, but I think I can prove mathematically that the discrete flooring will never change which type wins as well.

Referencing the showdown implementation:

```javascript
Math.floor(Math.floor(Math.floor(Math.floor(2 * pokemon.level / 5 + 2) * 90 * atk) / def) / 50);
```

The expression `Math.floor(2 * pokemon.level / 5 + 2)` will always be an integer after `Math.floor`.  If we simplify this to `C`:

```javascript
Math.floor(Math.floor(Math.floor(C * 90 * atk) / def) / 50);
```

C, 90 and atk are all integer values, so multiplying them will also always result in an integer.  That means the `Math.floor` call wrapping this is redundant, and can be dropped.

```javascript
Math.floor(Math.floor((C * 90 * atk) / def) / 50);
```

Multiplication is commutative, so you can drop those wrapping parens and see that the value of atk / def is computed before truncation, and so the resulting damage type depends on the exact ratio of atk / def.

---

With all that said, _IF_ the Smogon assumptions are accurate, the only remaining issue I can imagine is that sometimes the game might flatten a few close cases and randomly choose a type.  So we won't be _wrong_, but in some super rare case, we might make a different fixed guess than usual.

calc/src/mechanics/util.ts
```diff
@@ -356,6 +357,34 @@ export function getFinalDamage(
...
+ * Ratio seems more likely to me, since the pokemon damage formula is
+ * dependent on the ratio and not the difference of attack and defense stats.
+ *
+ * 2. In game, when stats are equal, category is random.  For damage, it won't
+ * matter most of the time (items like assault vest aren't included in this
+ * decision though, so *may* take effect). This implementation defaults to
+ * 'Special'.
+ *
+ * See also:
+ * {@link https://github.com/smogon/pokemon-showdown/commit/65d2bb5d0c36b61a2577382cc30aa861521113fd#diff-706ad2c9c5d1920e4245d5debe9e3c11}
```
scheibo, line 376:
```suggestion
 * {@link https://github.com/smogon/pokemon-showdown/commit/65d2bb5d}
```

calc/.eslintrc
scheibo: Would readily accept it into https://github.com/pkmn/eslint-config (I didn't know ignoreUrls was an option), though in this particular PR this style change is unnecessary :)

calc/src/mechanics/util.ts
urkerab: > Damage cannot be stochastic. You have two choices here:
> 
> - either simply favor one over the other and leave a comment explaining this decision = the approach ajhyndman originally went with
> - add a field to Move to configure how the tie should be broken (so that it could be wired up to a UI which allows the user to decide) = more correct, but potentially overkill

There is already a UI to override the category of a Move which you could use as your fallback.

> I suppose theres also hidden option c
> 
> - run the calculation twice with SSA being Physical once and Special in the other and report on both results

Ah, but the category doesn't depend on the whole calculation, only just this one part.

Marty-D:
The first research post @scheibo linked already proves it can't just be a simple comparison of (attack stat / defense stat). I make reference to those tests again in the last post he linked because this kind of thing didn't make any sense.
```
170 Atk / 59 Def = 2.881356
141 SpA / 49 SpD = 2.877551
```

Your implementation would have this always return Physical, but in reality it can be either Physical or Special at random.

ajhyndman:
> The first research post @scheibo linked already proves it can't just be a simple comparison of (attack stat / defense stat). I make reference to those tests again in the last post he linked because this kind of thing didn't make any sense.
> 
> ```
> 170 Atk / 59 Def = 2.881356
> 141 SpA / 49 SpD = 2.877551
> ```
> 
> Your implementation would have this always return Physical, but in reality it can be either Physical or Special at random.

Yeah, the PS implementation needs to preserve this coin flip behaviour, so I don't think it makes sense to optimize there.

However, as I mention above, for calculator purposes, it's desirable to always return a deterministic result, so I think this is fine.  As @urkerab mentions, there's already a UI in Showdown calc to override the category if you really need to.

calc/src/mechanics/util.ts
```diff
@@ -356,6 +357,34 @@ export function getFinalDamage(
...
+ * 1. As far as I can tell, research hasn't been tone on whether it's the larger
+ * ratio (atk / def) or the difference (atk - def) which decides the move
+ * category.
+ * Ratio seems more likely to me, since the pokemon damage formula is
+ * dependent on the ratio and not the difference of attack and defense stats.
+ *
+ * 2. In game, when stats are equal, category is random.  For damage, it won't
+ * matter most of the time (items like assault vest aren't included in this
+ * decision though, so *may* take effect). This implementation defaults to
+ * 'Special'.
```
scheibo, line 373:
```suggestion
 * A simplified formula can be used here compared to what the research 
 * suggests as we do not want to implement the random tiebreak element of
 * move - instead we simply default to 'Special' and allow the user to override
 * this by manually adjusting the move's category.
```

scheibo:
Thanks all! What a wack move this is :P

calc/src/mechanics/util.ts
```diff
@@ -356,6 +357,27 @@ export function getFinalDamage(
   return OF16(pokeRound(Math.max(1, OF32(damageAmount * finalMod) / 0x1000)));
 }
 
+/**
+ * Determines which move category Shell Side Arm should behave as.
+ *
+ * A simplified formula can be used here compared to what the research
```
scheibo, line 363:
```suggestion
 * A simplified formula can be used here compared to what the research
```

urkerab:
(I haven't actually looked to see how that UI override works, so I can't be sure that you're not actually overriding it rather than the other way around.)

scheibo:
The UI overrides `move.category`, so currently does not serve as a way for overriding the SSA in the calc.pokemonshowdown.com UI. That being said, it probably is the correct way to allow for overriding this (as opposed to adding yet another freaking toggle), so some rewiring is probably necessary here. That being said, allowing for users to break ties here seems like really low priority concern at the minute

PR #405, opened 2020-07-30
Add weights to all Gmax forms

ajhyndman:
Weight based moves don't work when dynamaxed — and this was [updated recently](https://github.com/smogon/damage-calc/commit/e3561c72a926c0972133b48427a8ae67573aad5e) (thanks @AustinXII).

However, weight based moves _should_ still work against the un-dynamaxed Gmax form of a pokemon.

### DIFF
--- a/calc/src/data/species.ts
+++ b/calc/src/data/species.ts
@@ -7503,7 +7503,7 @@ const SS_PATCH: {[name: string]: DeepPartial<SpeciesData>} = {
   'Alcremie-Gmax': {
     types: ['Fairy'],
     bs: {hp: 65, at: 60, df: 75, sa: 110, sd: 121, sp: 64},
-    weightkg: 0,
+    weightkg: 0.5,
     abilities: {0: 'Sweet Veil'},
     baseSpecies: 'Alcremie',
   },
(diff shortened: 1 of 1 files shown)


scheibo:
> **The data in calc/data must be kept in sync with Pokémon Showdown. If there is an issue with the calc's data, please first fix it in the simulator first.** In general, you should probably not be making manual edits to any of the data files, and in the future they are likely to be generated programmatically.

Pretty soon Gmax 'formes' will not even exist (**Gigantamax mons are not forms**), so this won't work even when we update the `@pkmn/dex` dependency. You're going to want to add logic here, not adjust the data files

ajhyndman:
Ohh.  My bad.

I'll keep this patch local then until the upstream changes go through!

scheibo:
^ should fix your bug and is what i meant by "You're going to want to add logic here, not adjust the data files", though its also not really the 'correct' approach until we change how we handle Gmax.

A new `@smogon/calc` package will be cut tomorrow or the next day when usage stats land and we update sets.

ajhyndman:
> ^ should fix your bug and is what i meant by "You're going to want to add logic here, not adjust the data files", though its also not really the 'correct' approach until we change how we handle Gmax.
> 
> A new `@smogon/calc` package will be cut tomorrow or the next day when usage stats land and we update sets.

Thanks!  I appreciate the workaround.  That'll save me some work if new gmax comes out before the upstream fix.

PR #407, opened 2020-08-31
Dauntless Shield is not a defense modifier

DaWoblefet:
It just raises Defense stage by 1, like using Defense Curl. It's certainly doesn't always boost it in Grassy Terrain, hahaha


scheibo:
Thanks!

PR #408, opened 2020-09-09
Implement Gigantamaxing properly

### LINKED ISSUE #406: Implement Gigantamax properly
Gigantamax formes and moves should not be included in the data files (they are not in the game data files, they are not formes/moves by that definition). PS has begun doing some work here, the most obvious being marking the 'formes'/'moves' as `Nonstandard` and adding `Gigantamax: Yes` to the importable. The following is all required:

- [ ] support `Gigantamax: Yes` in imports (moveset_import.js)
- [ ] remove nonstandard `Gigantamax` [from being special cased for inclusion](https://github.com/smogon/damage-calc/blob/master/calc/src/test/gen.ts#L491)
  - [ ] similarly, [clean up `otherFormes` hardcodes (in particular, Toxitricity/Urshifu)](https://github.com/smogon/damage-calc/blob/master/calc/src/test/gen.ts#L312-L342) to remove the Gmax 'forme'
- [ ] Gigantamax should be a checkbox in the UI instead of a 'forme'.
- [ ] Set import logic should switch to sets with `gigantmax: true` instead of the legacy `-Gmax` name approach

The main question is around moves - how should the Gmax version of a move be handled? **This should probably block on PS implementing the Gmax data as a field on the base move**.

KrisXV:
Closes https://github.com/smogon/damage-calc/issues/406

This could certainly use some UI improvements @AustinXII and iirc you're more familiar with the UI so I'll let you take a look


AustinXII:
What does the UI look like currently with this?

AustinXII:
@KrisXV update on this?

PR #412, opened 2020-10-02
Double Facade base power when poisoned, paralyzed or burned

Ludenife:
Missing effect in gen 3.


Ludenife:
Thought this was a simple fix, should have opened an issue instead, my bad.

AustinXII:
this fix is as simple as this, the build failure isn't because of you.

PR #417, opened 2020-12-16
fixing the export bug

elad-m:
I redid it, and it looks better now (only 1 commit) I hope. I closed the previous PR.


KrisXV:
Thanks, and sorry for the delay from before!

PR #421, opened 2021-01-08
adding dockerfile

unsupo:
simple dockerfile which builds the project and serves the static webpage via nginx

Either this or #422

The other one hosts an api and doesn't use nginx but adds a new dependency

https://docs.docker.com/get-docker/
The following checks if docker image is running and stops it, then builds the image and lastly starts it on port 1000
```bash
name=damage_calc
img=unsupo/damage-calc
docker ps -a | grep ${name} > /dev/null 2>&1 && { docker stop ${name} > /dev/null 2>&1; docker rm ${name} > /dev/null 2>&1; }
docker build -t ${img} .
docker run -p 1000:80 --name ${name} -d ${img}
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,23 +1,24 @@
 FROM node:12 as build
 # Create app directory
-WORKDIR /usr/src/app
+WORKDIR /usr/src/app/calc
 
 # Install app dependencies
 # A wildcard is used to ensure both package.json AND package-lock.json are copied
 # where available (npm@5+)
+COPY calc/package*.json ./
+
+RUN npm install
+# If you are building your code for production
+# RUN npm ci --only=production
+WORKDIR /usr/src/app
+
 COPY package*.json ./
 
 RUN npm install
 
-WORKDIR /usr/src/app/calc
-COPY calc/package*.json ./
-RUN npm install
-# If you are building your code for production
-# RUN npm ci --only=production
-
 # Bundle app source
+COPY package*.json
 
-WORKDIR /usr/src/app
 COPY . .
 RUN node build
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,28 @@
+FROM node:12 as build
+# Create app directory
+WORKDIR /usr/src/app
+
+# Install app dependencies
+# A wildcard is used to ensure both package.json AND package-lock.json are copied
+# where available (npm@5+)
+COPY package*.json ./
+
+RUN npm install
+
+WORKDIR /usr/src/app/calc
+COPY calc/package*.json ./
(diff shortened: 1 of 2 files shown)


unsupo:
@KrisXV i've added instructions to get this to work.

Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM node:12 as build
+# Create app directory
+WORKDIR /usr/src/app
+
+# Install app dependencies
+# A wildcard is used to ensure both package.json AND package-lock.json are copied
+# where available (npm@5+)
+COPY package*.json ./
+
+RUN npm install
```
aviettran, line 10: You need to run both COPY commands before doing the first install, as the package.json depend on each other.

aviettran (commented):
> docker run -p 1000:3000 --name ${name} -d ${img}

nginx listens on 80 by default, not 3000

Dockerfile
unsupo: If you look i copy and install in this directory and then i copy and install the calc directory then i build since it depends on the calc dir.
This is the same instructions from the doc and it works

unsupo:
> > docker run -p 1000:3000 --name ${name} -d ${img}
> 
> nginx listens on 80 by default, not 3000

You are right, i did neglect to port forward the right port in my example script here, but if you pull the prs you'll see they do in fact work.

Dockerfile
unsupo: Yes, this would get around the no package-lock.json committed.
Why is package-lock.json being ignore in the .gitignore?

unsupo:
@aviettran I modified it so it works without package-lock.json.  Thank you for the comments

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+# If you are building your code for production
+# RUN npm ci --only=production
+WORKDIR /usr/src/app
+
+COPY package*.json ./
+
+RUN npm install
+
+# Bundle app source
+COPY package*.json
```
aviettran, line 20: should be `COPY package*.json ./`

unsupo: Yes it should

unsupo:
@aviettran yeah don't know how i missed the dest path.  Thanks

aviettran:
docker functionality working as expected

AustinXII:
@aviettran which docerfile would you prefer? Sorry none of the members I have asked have used dockerfiles before

aviettran:
The Dockerfiles are very similar and only differ in how they serve http (nginx vs node/express). The API PR adds express routes (an API), however, so that should be the basis by which you decide which PR is more appropriate.

PR #422, opened 2021-01-08
Docker addition api

unsupo:
Either this or https://github.com/smogon/damage-calc/pull/421

This one hosts an api and doesn't use nginx but adds a new dependency

https://docs.docker.com/get-docker/
The following checks if docker image is running and stops it, then builds the image and lastly starts it on port 1000
```bash
name=damage_calc
img=unsupo/damage-calc
docker ps -a | grep ${name} > /dev/null 2>&1 && { docker stop ${name} > /dev/null 2>&1; docker rm ${name} > /dev/null 2>&1; }
docker build -t ${img} .
docker run -p 1000:3000 --name ${name} -d ${img}
```


unsupo:
@KrisXV i've added instructions to get this to work.

aviettran:
See my comment for #421, as this branch has the same issue.

unsupo:
> See my comment for #421, as this branch has the same issue.
@aviettran 

How could this one have "the same issue"?  Your comment on the other one was that nginx uses port 80 by default this one clearly exposes port 3000 in the dockerfile and in server.js file and doesn't use nginx

Also in regards to your other comment the fun part of docker is you can map a port to a different port, so maybe my example copy and paste script might have been wrong for the other one, what is your point on this one?

aviettran:
I was referring to the other comment regarding the package.json

unsupo:
> I was referring to the other comment regarding the package.json

The only issue with package.json is this repo ignores the package-lock.json with one in place it fixes the dependency issue
but i could modify the dockerfile if the package-lock.json can't be added for whatever reason. 
Thanks for your comments @aviettran

aviettran:
You can add a lock to the repo, but the image should still build without it. The presence of the lock just masks the issue. If you copy both `package.json` first, the dependency issue is resolved and it will build even without the lock.

unsupo:
@aviettran I've made the requested fix, please review.  I do prefer this one was it exposes an api, but the #421 would work fine if you don't want to add the express dev dependency

aviettran:
docker functionality working as expected

AustinXII:
Sorry for the late reply! This was something @KrisXV asked for?

Issue #427, opened 2021-01-16
Search doesn't work when using "Only show imported sets" mode


brunopugliese:
When trying to search for a Pokémon with the "Only show imported sets" option checked, it does not work at all.

PR #428, opened 2021-01-16
Search for imported sets

brunopugliese:
Replicated the search capabilities of the default list to the "only show imported sets" filtered list.
It now also searches for set name and pokémon name instead of just pokémon name in that filtered mode.

Fixes Issue #427


AustinXII:
Geeze how did I miss this? Thanks!!

PR #430, opened 2021-02-06
Support field-wide Fairy Aura, Dark Aura and Aura Break when neither Pokemon uses the ability

GriffinLedingham:
Fairy Aura, Dark Aura, and Aura Break currently rely on one of the two Pokemon involved in the calc using the appropriate ability. To better support partners carrying these abilities in VGC/doubles, these three ability effects have been added to the Field class.

The missing support for this functionality was brought to me by @jbmagier and I arrived at a solution following some discussion between us.

This PR simply implements the under the hood functionality for use in the @smogon/calc module, and does not implement the front end support. This functionality can be seen in use on the Pikalytics Damage Calc in the screenshots below:

Screen Shot 2021-02-06 at 11 00 30 AM
Screen Shot 2021-02-06 at 11 00 14 AM

### DIFF
--- a/calc/src/field.ts
+++ b/calc/src/field.ts
@@ -6,6 +6,9 @@ export class Field implements State.Field {
   weather?: Weather;
   terrain?: Terrain;
   isGravity: boolean;
+  isAuraBreak?: boolean;
+  isFairyAura?: boolean;
+  isDarkAura?: boolean;
   attackerSide: Side;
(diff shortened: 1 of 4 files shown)


GriffinLedingham:
Reorganized a portion of the Aura conditionals for clarity and to reduce line length as per eslint.

AustinXII:
thanks!

PR #433, opened 2021-03-02
Fix Reversal and Flail power calculation for Gen 4

wartab:
After investigation in Generation 4 speedrunning (thanks minnow), it appears that the thresholds for Gen 4 are wrong when calculating the power of Flail and Reversal.

The calculation is correct for Gen 5+, though.

### DIFF
--- a/calc/src/mechanics/gen4.ts
+++ b/calc/src/mechanics/gen4.ts
@@ -162,8 +162,8 @@ export function calculateDPP(
     break;
   case 'Flail':
   case 'Reversal':
-    const p = Math.floor((48 * attacker.curHP()) / attacker.maxHP());
-    basePower = p <= 1 ? 200 : p <= 4 ? 150 : p <= 9 ? 100 : p <= 16 ? 80 : p <= 32 ? 40 : 20;
+    const p = Math.floor((64 * attacker.curHP()) / attacker.maxHP());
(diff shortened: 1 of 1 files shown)


AustinXII:
can you fix travis?

DaWoblefet:
Kris asked me to verify if this is actually the case. I loaded up my Gen 4 game to see (ENG SoulSilver vs. ENG Platinum; host SoulSilver) and had a Blissey with 23/714 HP use Flail on another target. If Flail were using a constant of 48, this should qualify as a BP of 200, because floor(23*48 / 714) = 1. If Flail was using a constant of 64, this should qualify as a BP of 150, because floor(23*64 / 714) = 2. Given this level 100 Blissey with 25 Attack and a target Dewgong with 185 Defense, I dealt 24 and 25 damage. So this gets my seal of approval!

As an aside, I find this really bizarre; the [Emerald disassembly](https://github.com/pret/pokeemerald/blob/master/src/battle_script_commands.c) uses identical scaling to both [Gen 5](https://www.smogon.com/bw/articles/bw_complete_damage_formula#base-power)) and Gen 7. I always thought the reason they used 48 was because there was 48 pixels on the HP bar - I wonder if that wasn't the case in Gen 4.

KrisXV:
Thanks for this!

PR #440, opened 2021-04-07
Fix Rivalry boost

Flametix:
https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/page-8#post-8199267 and https://www.smogon.com/bw/articles/bw_complete_damage_formula agree on 0x1400 and 0xC00. I'm not sure about gen 4 but it should be the same.


DaWoblefet:
Verifying this change is accurate.

KrisXV:
Thanks! Sorry for the delay on merging; various changes got in the way, and I wanted to verify this.

PR #444, opened 2021-04-27
Convert modifiers from hex to decimal

DaWoblefet:
See similar PR done on Showdown main server; https://github.com/smogon/pokemon-showdown/pull/8003

This helps for maintenance so you don't have to pull out a calculator to convert back and forth. I don't claim that all modifiers here are accurate, but all conversions should be accurate. I did make one change beyond conversions, since I noticed it was low-hanging fruit: Auras in Gen 6 should also be 5448.


KrisXV:
thx

PR #445, opened 2021-05-02
Add FFA spread move modifier support

DaWoblefet:
In order to not have doubles look rounded off in past gens, I added two labels for it, one that only displays in Gen 8 (with the btn-middle class) and one that only displays in Gens 3-7 (with the btn-right class). I am pretty sure this is not a super good way to do it (you see the extra label on page load). The proper way sounds to me like making the group of buttons into a flexbox and just applying the styling on first and last childs, but that would be a decent amount of CSS refactoring.

Screenshots:
image

image

image

### DIFF
--- a/calc/src/data/interface.ts
+++ b/calc/src/data/interface.ts
@@ -12,7 +12,7 @@ export type SpeciesName = string & As<'SpeciesName'>;
 
 export type StatusName = 'slp' | 'psn' | 'brn' | 'frz' | 'par' | 'tox';
 
-export type GameType = 'Singles' | 'Doubles';
+export type GameType = 'Singles' | 'Doubles' | 'FFA';
 export type Terrain = 'Electric' | 'Grassy' | 'Psychic' | 'Misty';
 export type Weather =
   | 'Sand' | 'Sun' | 'Rain' | 'Hail' | 'Harsh Sunshine' | 'Heavy Rain' | 'Strong Winds';
--- a/calc/src/data/interface.ts
+++ b/calc/src/data/interface.ts
@@ -12,7 +12,7 @@ export type SpeciesName = string & As<'SpeciesName'>;
 
 export type StatusName = 'slp' | 'psn' | 'brn' | 'frz' | 'par' | 'tox';
 
-export type GameType = 'Singles' | 'Doubles' | 'FFA';
+export type GameType = 'Singles' | 'Doubles' | 'Free-For-All';
 export type Terrain = 'Electric' | 'Grassy' | 'Psychic' | 'Misty';
 export type Weather =
   | 'Sand' | 'Sun' | 'Rain' | 'Hail' | 'Harsh Sunshine' | 'Heavy Rain' | 'Strong Winds';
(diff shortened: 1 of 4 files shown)


DaWoblefet:
OK I played around with this for about an hour and refactoring this to not use 2 labels would require some serious refactors to keep the styling consistent as before (because we are just styling away elements with certain classes rather than removing them from the DOM).

calc/src/data/interface.ts
```diff
@@ -12,7 +12,7 @@ export type SpeciesName = string & As<'SpeciesName'>;
 
 export type StatusName = 'slp' | 'psn' | 'brn' | 'frz' | 'par' | 'tox';
 
-export type GameType = 'Singles' | 'Doubles';
+export type GameType = 'Singles' | 'Doubles' | 'FFA';
```
KrisXV, line 15: Should be Free-For-All

src/index.template.html
```diff
@@ -631,12 +631,15 @@
...
+                <div aria-labelledby="selectBattleFormatInstruction" class="gen-specific g3 g4 g5 g6 g7 g8" role="radiogroup" style="margin: 0 auto 5px;" title="Select the battle format.">
                     <span class="visually-hidden" id="selectBattleFormatInstruction">Select the battle format.</span>
                     <input class="visually-hidden calc-trigger" type="radio" name="format" value="Singles" id="singles-format" checked="checked" />
                     <label class="btn btn-left" for="singles-format">Singles</label>
                     <input class="visually-hidden calc-trigger" type="radio" name="format" value="Doubles" id="doubles-format" />
-                    <label class="btn btn-right" for="doubles-format">Doubles</label>
+                    <label class="btn btn-right gen-specific g3 g4 g5 g6 g7" for="doubles-format">Doubles</label>
+                    <label class="btn btn-mid gen-specific g8" for="doubles-format">Doubles</label>
+                    <input class="visually-hidden calc-trigger gen-specific g8" type="radio" name="format" value="FFA" id="ffa-format" />
```
KrisXV, line 641: ditto

src/randoms.template.html
```diff
@@ -645,12 +645,15 @@
...
+                <div aria-labelledby="selectBattleFormatInstruction" class="gen-specific g3 g4 g5 g6 g7 g8" role="radiogroup" style="margin: 0 auto 5px;" title="Select the battle format.">
                     <span class="visually-hidden" id="selectBattleFormatInstruction">Select the battle format.</span>
                     <input class="visually-hidden calc-trigger" type="radio" name="format" value="Singles" id="singles-format" checked="checked" />
                     <label class="btn btn-left" for="singles-format">Singles</label>
                     <input class="visually-hidden calc-trigger" type="radio" name="format" value="Doubles" id="doubles-format" />
-                    <label class="btn btn-right" for="doubles-format">Doubles</label>
+                    <label class="btn btn-right gen-specific g3 g4 g5 g6 g7" for="doubles-format">Doubles</label>
+                    <label class="btn btn-mid gen-specific g8" for="doubles-format">Doubles</label>
+                    <input class="visually-hidden calc-trigger gen-specific g8" type="radio" name="format" value="FFA" id="ffa-format" />
```
KrisXV, line 655: ditto

PR #439, opened 2021-03-30
Function modularization tweaks (gens 7, 8) and iron ball calculations

lighthouse64:
Iron ball calculations apply to all generations.  This commit includes new test cases for iron ball calculations as well.

TravisCI errors are coming from the two problems that existed prior to me pulling.

PR #446, opened 2021-05-02
Gen 7/8: Fix -ate abilities, Reckless, and Iron Fist

DaWoblefet:
Something about https://github.com/smogon/damage-calc/pull/439 changed how -ate abilities worked with typing; currently, whenever the move enters the function, it is already Fairy / Electric / etc. and so the check for "if you're a Normal-type move" doesn't work. To fix this, I create a new move object to check if the original move really is Normal. I'm not super sure why this worked before; probably the correct solution is to set a flag or something.

Splitting off the if statement on Reckless and Iron Fist fixes that part.

Fixes:
https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/post-8841109
https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/post-8841181

### DIFF
--- a/calc/src/mechanics/gen78.ts
+++ b/calc/src/mechanics/gen78.ts
@@ -825,17 +825,20 @@ export function calculateBPModsSMSS(
   );
 
   if (!move.isZ && !move.isMax && !noTypeChange) {
-    const normal = move.hasType('Normal');
-    if (attacker.hasAbility('Aerilate') && normal ||
-      attacker.hasAbility('Galvanize') && normal ||
-      attacker.hasAbility('Pixilate') && normal ||
-      attacker.hasAbility('Refrigerate') && normal ||
-      attacker.hasAbility('Normalize')
+    // The -ate abilities already changed move typing earlier
+    const normal = new Move(gen, move.originalName).hasType('Normal');
+    if (attacker.hasAbility('Normalize') ||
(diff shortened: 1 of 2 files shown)


DaWoblefet:
I have absolutely no idea what this build error is.

calc/src/mechanics/gen78.ts
```diff
@@ -825,17 +825,20 @@ export function calculateBPModsSMSS(
...
 
   if (!move.isZ && !move.isMax && !noTypeChange) {
-    const normal = move.hasType('Normal');
-    if (attacker.hasAbility('Aerilate') && normal ||
-      attacker.hasAbility('Galvanize') && normal ||
-      attacker.hasAbility('Pixilate') && normal ||
-      attacker.hasAbility('Refrigerate') && normal ||
-      attacker.hasAbility('Normalize')
+    // The -ate abilities already changed move typing earlier
+    const normal = new Move(gen, move.originalName).hasType('Normal');
```
DaWoblefet, line 829: Build error seems to be from this line. Guess I'll need to find a new way to check if the move was originally Normal-type.

DaWoblefet:
I resolved this by storing a flag if an ate Ability should boost the user. This is closer to how the games do it as well.

KrisXV:
thank u

Issue #443, opened 2021-04-25
Current mechanics inaccuracies [WIP]


DaWoblefet:
I am doing a static analysis of the damage formula in Gen 7 / 8. Currently work-in-progress. I will continue editing as I find more. I will assist in fixing these!

Reference for how modifiers are actually chained: https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/post-8685465

### Weight
- Light Metal and Float Stone should quarter speed.
- Weight multiplication is done with integers hectograms, not kilograms (i.e. you can't have a weight of 0.15kg). It also should round.

### Moves with custom base power
- Some moves have hardcoded BP instead of move.bp; these should be swapped to move.bp and move.bp * 2, etc. Payback, Acrobatics, Weather Ball
- Some moves attempt to code in move failures by setting BP to 0; it should be more like weight moves into Dynamax instead of setting BP to 0. Poltergeist, Steel Roller
- Expanding Force is a BP modifier, not a move with variable base power.
- Misty Explosion is a BP modifier, not a move with variable base power. It should be checking for groundedness.
- Rising Voltage is a BP modifier, not a move with variable base power. [probably, still need to prove]
- Gyro Ball's formula is wrong; should be 1 + floor((25*target Speed) / user speed)
- Nature Power should probably add a desc indicating what move it transformed into and not hardcode BP
- Triple Axel hack is hack (but may as well be applied to Triple Kick while you're there)
- Crush Grip / Wring Out implementation could be refactored for readability; I think it's technically correct
- Smelling Salts into paralyzed target is missing, should be easy to add

### BP modifiers
Generally speaking these should be applied in the following order, with some Speed checks (see link at top of issue):
1. Move effects
2. Field effects
3. Abilities
4. Items

- Technician checks if a move is 60 or less after it evaluates moves with custom base power. It doesn't consider any base power modifiers. This was mistaken in Gen 7, it works the same way in 7 & 8.
- Auras are dichotomous and applied in the same step. It's 5448 with Aura, 3072 with Aura Break; they aren't two separate modifiers.
- Rivalry is very wrong but a fix is pending in https://github.com/smogon/damage-calc/pull/440
- Power Spot can be supported, 5325, pending in https://github.com/smogon/damage-calc/pull/435

### Attack modifiers
Generally speaking these should be applied in the following order, with some Speed checks (see link at top of issue):
1. Abilities
2. Items

- Hustle should be a standard 6144 / 4096
[...]

PR #447, opened 2021-05-11
Improve base power modifiers

DaWoblefet:
Implements most of the base power issues in https://github.com/smogon/damage-calc/issues/443; I didn't want to do everything at once so it'd be easier to code review. Notably, I did not implement the Speed-based modifier chaining yet; I'll need to think about that some more and want to tackle it in a separate PR once I take care of the low-hanging fruit.

Moves with custom BP:
- Steel Roller and Poltergeist now fail earlier before even entering BP calculation
- Payback, Acrobatics, Weather Ball now no longer use hardcoded damage values
- Pursuit was changed to be a move with variable base power
- Smelling Salts and Triple Kick were added as moves with variable base power
- Gyro Ball's formula was corrected

Base Power modifiers:
- BP modifiers are now grouped by effect type (move, ability, etc.) then by modifier
- Zamazenta with Rusted Shield doesn't increase Knock Off
- Expanding Force, Rising Voltage, and Misty Explosion changed to base power modifiers
- Technician's mechanics were corrected (the past theory about temporary chaining is very wrong)
- Aura mechanics were corrected (they only apply once)
- Punk Rock (offensive) changed to a base power modifier

Misc:
- Terrain Pulse and Nature Power now list their typing and base power in desc

### DIFF
--- a/calc/src/mechanics/gen56.ts
+++ b/calc/src/mechanics/gen56.ts
@@ -262,7 +262,13 @@ export function calculateBWXY(
 
   switch (move.name) {
   case 'Payback':
-    basePower = turnOrder === 'last' ? 100 : 50;
+    basePower = move.bp * (turnOrder === 'last' ? 2 : 1);
+    desc.moveBP = basePower;
+    break;
+  case 'Pursuit':
+    const switching = field.defenderSide.isSwitching === 'out';
+    basePower = move.bp * (switching ? 2 : 1);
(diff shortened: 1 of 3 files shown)


DaWoblefet:
Another bump for review @KrisXV - I wanted to get the entirety of https://github.com/smogon/damage-calc/issues/443 done over the summer. I'll merge myself if you're still busy, since these are objective bugfixes.

KrisXV:
thx

PR #449, opened 2021-06-11
Karate chop should crit

BenjaminShinar:
I've discovered that Karate Chop didn't always crit for Gen 1.


KrisXV:
Thanks for catching this!

PR #453, opened 2021-07-06
Fix Dauntless Shield

Karthik99999:
The Dauntless Shield check was being run after the final defense stats were calculated. I don't understand why `computeFinalStats` has to be run separately for defense and attack anyway, but yeah this is why it was bugged the whole time.


Karthik99999:
this seems to have been fixed here https://github.com/smogon/damage-calc/commit/8d527cb11ed58bb19a3302ed093aa9495a07d3a6 but somehow got reverted? not sure how that happened, considering other stuff from that commit still exists.

KrisXV:
I'm unsure how it got reverted; either way, thanks!

PR #455, opened 2021-07-07
Fix Wonder Room

Karthik99999:
apparently applying it in `computeFinalStats` doesn't work (probably should've tested that before i added it lol)


DaWoblefet:
Thanks Karthik!

PR #457, opened 2021-08-09
Nature Power fixes

Karthik99999:
- Correctly adjusts move category
- Uses correct moves for gens 3 and 5 (swift and earthquake respectively)
- Changes the move name in the calc description to match the move being used (so Thunderbolt will show up in the calc description if Nature Power is selected in Electric Terrain)


KrisXV:
looks good aside from the conflicts, fix those and ill merge

PR #458, opened 2021-08-19
Triple Axel/Triple Kick fixes

Karthik99999:
Made Triple Axel/Triple Kick have a multihit property between 1-3, and implemented Triple Axel's bp hack for Triple Kick. I also made it so that the dropdown to select how many hits a multihit move will hit correctly changes the options from 2-5 hits to 1-3 hits when using Triple Axel/Triple Kick.


DaWoblefet:
Although the UI is updated, it doesn't seem to actually be influencing damage calculation. Neither the fake estimated damage roll nor the description is updated.
image

calc/src/move.ts
```diff
@@ -89,14 +89,18 @@ export class Move implements State.Move {
...
         } else {
-          this.hits = (options.ability === 'Skill Link' || options.item === 'Grip Claw')
-            ? data.multihit[1]
-            : data.multihit[0] + 1;
+          if (typeof data.multihit === 'number') {
+            this.hits = data.multihit;
+          } else if (options.hits) {
+            this.hits = options.hits;
+          } else {
+            this.hits = (options.ability === 'Skill Link' || options.item === 'Grip Claw')
```
DaWoblefet, line 100: Unrelated to the PR, but I have no idea what Grip Claw is doing here; it doesn't have any effect on multihit moves. Looks like a very old mechanics inaccuracy introduced a [few years ago](https://github.com/smogon/damage-calc/commit/1f95cd0bc456ed2903b9d578a391fbcd3feeb52d). I'll correct it.

AustinXII: resolved in #510

Zrp200:
would it be possible to account for loaded dice as well?

Karthik99999:
> would it be possible to account for loaded dice as well?

I don't think this is needed since the number of hits is being manually set

PR #459, opened 2021-08-21
Fix various Speed mechanics

DaWoblefet:
This implements a number of fixes to calculating Speed, including:

- Speed now has a proper floor of 0; the Gyro Ball and Electro Ball edge cases are implemented to reflect this
- Speed now uses proper modifier chaining
- Lower / upper bounds are implemented on all chained modifiers; in practice, this will only end up mattering for Speed
- Applying stat boosts / drops for stat calculation in Gen 3+ now accounts for 16-bit overflow


AnnikaCodes:
Thanks!

PR #469, opened 2022-02-06
Fix Expanding Force in doubles

Karthik99999:
Spread damage reduction was not being applied to expanding force when the field was set to doubles


DaWoblefet:
Build failure looks unrelated to this PR. Thanks Karthik!

PR #470, opened 2022-03-09
Ripen not applied to super effective berries

mmorrisonlk:
I was messing around testing something on Flapple. I noticed the damage calculation was the same on ice moves with a Yache Berry held with and without Ripen as the ability. Ripen should double the effect of all berries. So a super effective attack should deal 1/4th damage and not half damage. I believe this should fix that case assuming I understood the surrounding code in regards to the damage formula.


AnnikaCodes:
Thanks!

PR #476, opened 2022-07-22
Fix Technician's interaction with custom BP moves

sparksblade:
Using basePower instead of move.bp to correctly apply (and not apply) technician for moves with custom bp


DaWoblefet:
Mechanics are correct here. Thanks!

PR #485, opened 2022-11-20
Add Terastallization support

ajhyndman:
- Adds teraType to Pokemon class
- Accounts for teraType in STAB calculation
- Accounts for teraType in resistance calculation

You might want to double-check especially the resistance math.  I am not confident I captured all edge cases there.

### DIFF
--- a/calc/src/mechanics/gen78.ts
+++ b/calc/src/mechanics/gen78.ts
@@ -231,25 +231,25 @@ export function calculateSMSS(
   }
 
   const isGhostRevealed = attacker.hasAbility('Scrappy') || field.defenderSide.isForesight;
+  const isRingTarget = defender.hasItem('Ring Target') && !defender.hasAbility('Klutz');
   const type1Effectiveness =
-    getMoveEffectiveness(gen, move, defender.types[0], isGhostRevealed, field.isGravity);
+    getMoveEffectiveness(gen, move, defender.types[0], isGhostRevealed, field.isGravity, isRingTarget);
   const type2Effectiveness = defender.types[1]
-    ? getMoveEffectiveness(gen, move, defender.types[1], isGhostRevealed, field.isGravity)
+    ? getMoveEffectiveness(gen, move, defender.types[1], isGhostRevealed, field.isGravity, isRingTarget)
     : 1;
   let typeEffectiveness = type1Effectiveness * type2Effectiveness;
 
-  if (typeEffectiveness === 0 && move.named('Thousand Arrows')) {
-    typeEffectiveness = 1;
-  } else if (typeEffectiveness === 0 && move.hasType('Ground') &&
+  if (defender.teraType) {
+    typeEffectiveness = getMoveEffectiveness(gen, move, defender.teraType, isGhostRevealed, field.isGravity, isRingTarget);
+  }
+
+  if (typeEffectiveness === 0 && move.hasType('Ground') &&
     defender.hasItem('Iron Ball') && !defender.hasAbility('Klutz')) {
     typeEffectiveness = 1;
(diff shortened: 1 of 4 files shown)


ajhyndman:
It looks like I broke a couple tests.  I'm investigating.

ajhyndman:
Okay, think I got it.  I didn't recall that iron ball makes effectiveness exactly 1x regardless of second type!

ajhyndman:
Closed in favor of #484

ajhyndman:
@AustinXII I think I got tests passing whenever you have a chance to take a look!

calc/src/mechanics/gen78.ts
```diff
@@ -230,26 +235,48 @@ export function calculateSMSS(
...
+  );
   const type2Effectiveness = defender.types[1]
-    ? getMoveEffectiveness(gen, move, defender.types[1], isGhostRevealed, field.isGravity)
+    ? getMoveEffectiveness(
+      gen,
+      move,
+      defender.types[1],
+      isGhostRevealed,
+      field.isGravity,
+      isRingTarget
```
KrisXV, line 257:
```suggestion
        gen,
        move,
        defender.types[1],
        isGhostRevealed,
        field.isGravity,
        isRingTarget
```

calc/src/mechanics/gen78.ts
```diff
@@ -455,16 +482,18 @@ export function calculateSMSS(
...
-    } else {
-      stabMod = 6144;
-    }
+    stabMod += 2048;
   } else if (attacker.hasAbility('Protean', 'Libero')) {
-    stabMod = 6144;
+    stabMod += 2048;
     desc.attackerAbility = attacker.ability;
   }
+  if (attacker.teraType === move.type) {
```
KrisXV, line 490: The stabBoost only gets an additional 4/3 multiplier if the base type of the Pokemon matches the tera type

calc/src/mechanics/gen78.ts
```diff
@@ -388,7 +415,7 @@ export function calculateSMSS(
   // #region (Special) Attack
   const attack = calculateAttackSMSS(gen, attacker, defender, move, field, desc, isCritical);
   const attackSource = move.named('Foul Play') ? defender : attacker;
-  if (move.named('Photon Geyser', 'Light That Burns The Sky')) {
+  if (move.named('Photon Geyser', 'Light That Burns The Sky', 'Tera Blast')) {
```
KrisXV, line 418: tera blast only does the category adjustment if the user is terastallized

calc/src/mechanics/gen78.ts
ajhyndman: The lint rule configured on this repo actually rejects that indentation.

KrisXV: That's weird, it shouldn't be on the same indentation as the first line

calc/src/mechanics/gen78.ts
ajhyndman: That's true, but I've organized this so that normal STAB and tera STAB stack additively, handling both cases.

If you check the math, I think you'll find it works out.

calc/src/mechanics/gen78.ts
ajhyndman: Whoah, I didn't actually know that.  Thanks for the catch!

calc/src/mechanics/gen78.ts
ajhyndman: Yeah, it looked wrong to me too.  🤷   You could always switch to prettier!

AnnikaCodes:
Thanks!

PR #486, opened 2022-11-24
Misc. developer experience tweaks

ajhyndman:
I just figured I would suggest a couple minor developer workflow suggestions.

1. Remove the "pretest" build step.  Jest is configured to run against the typescript source directly, so this looks unnecessary.  Removing this makes re-running tests and getting feedback much quicker.
2. Commit npm's `package-lock.json` files.  This is useful because it ensures that transitive dependencies (_dependencies of your package's dependencies_, e.g. `@pkmn/dex`) get the same version installed for every developer working on the package.  This helps me avoid having to delete my `node_modules` folder and re-installing every time I switch branches.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,3 @@ eslint-cache/
 Thumbs.db
 pnpm-debug.log
 npm-debug.log
-package-lock.json
--- a/calc/.gitignore
+++ b/calc/.gitignore
@@ -6,5 +6,4 @@ coverage/
 .DS_Store
 Thumbs.db
 npm-debug.log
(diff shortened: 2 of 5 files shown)


AnnikaCodes:
Please re-gitignore the package-lock file. @smogon policy is to not commit it; an explanation can be found [https://github.com/smogon/pokemon-showdown/blob/master/CONTRIBUTING.md#package-lockjson](here).

ajhyndman:
Fair enough!  I just thought I would make the suggestion.  I'll close the PR.

PR #487, opened 2022-11-24
Add Hisui forms to species data

ajhyndman:
Adds Hisui form pokemon to the dex.

It's not clear to me which generation these pokemon belong in, but considering that every one of these pokemon lines is in Scarlet/Violet, it seems highly likely that they will be added at some point, so it would be nice to have them in the calculator.

Given the test failures, I think I should propose this change to @pkmn/dex first.  Does that sound right?

### DIFF
--- a/calc/src/data/species.ts
+++ b/calc/src/data/species.ts
@@ -8756,7 +8756,13 @@ const SS_PATCH: {[name: string]: DeepPartial<SpeciesData>} = {
     types: ['Water', 'Ghost'],
     bs: {hp: 120, at: 112, df: 65, sa: 80, sd: 75, sp: 78},
     weightkg: 110,
-    abilities: {0: 'Rattled'},
+    abilities: {0: 'Swift Swim'},
+  },
+  'Basculegion-F': {
+    types: ['Water', 'Ghost'],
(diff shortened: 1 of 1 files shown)


AnnikaCodes:
I'm pretty sure the Hisui formes are already in @pkmn/dex, the calc just needs to be updated to recognize them as legal.

ajhyndman:
@AnnikaCodes I think it would be great to refactor this library to read species directly from @pkmn/dex, but right now that does not appear to be the case.

As far as I can tell, @pkmn/dex is only used for test assertions about the data specified here.

leo60228:
Basculegion's base stats depend on its gender, which doesn't seem to be accounted for in this PR.

ajhyndman:
@leo60228 thanks for the catch!  updated

leo60228:
The test failure is because the Hisui forms have `isNonstandard: 'future'` in `@pkmn/dex`, which makes this line exclude them: https://github.com/smogon/damage-calc/blob/master/calc/src/test/gen.ts#L496

KrisXV:
Hi, I ended up doing this in https://github.com/smogon/damage-calc/commit/21b84e1da6f58d38c3331a7bc5be9028b89132f0, but I appreciate your pull request and hope you'll still help out in the future!

PR #490, opened 2022-11-24
Missing/incomplete move data

ajhyndman:
- Flower Trick gives a guaranteed critical hit
- Wave Crash was missing from moves


AnnikaCodes:
Thanks!

PR #492, opened 2022-11-25
Population Bomb should not be guaranteed to hit 10 times

dragonakout:
According to both [bubapedia](https://bulbapedia.bulbagarden.net/wiki/Population_Bomb_(move)), [serebii](https://www.serebii.net/attackdex-sv/populationbomb.shtml) and in-game experience, Population Bomb is a multi-hit move that has a 1 to 10 chances of hitting an opponent and chances to miss. 

Not guaranteed 10 hits. 

I am not familiar with the project to fix the teambuilder description, but it should be updated as well.

### DIFF
--- a/calc/src/data/moves.ts
+++ b/calc/src/data/moves.ts
@@ -4395,7 +4395,7 @@ const SV_PATCH: {[name: string]: DeepPartial<MoveData>} = {
     zp: 100,
     maxPower: 90,
     makesContact: true,
-    multihit: 10,
+    multihit: [1-10],
   },
   Pounce: {
     bp: 50,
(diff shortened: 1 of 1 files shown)


ajhyndman:
@dragonakout I am not a maintainer of this project, but I think this move is already modelled correctly, actually.

This move has the same logic as Triple Kick and Triple Axel.  The number of hits is not rolled, it is always set to 10.  But accuracy is checked on each hit and the first miss will stop the move.

See Triple Kick in data:

https://github.com/smogon/damage-calc/blob/master/calc/src/data/moves.ts#L3605

---

With the above said, reading between the lines, I wonder if you are trying to provide feedback to the Pokemon Showdown app?  If that's the case, you might be looking for this repository instead!

https://github.com/pkmn/ps

dragonakout:
The wording made me think the implementation was wrong, but you are right @ajhyndman, the move always hits 10 times. Thank you for the kind response and have a good day

PR #495, opened 2022-11-28
Add support for new abilities and snow weather

ajhyndman:
Implements the following:

- Snow increases physical defense in generation 9
- Protosynthesis & Quark Drive
- The four Treasures of Ruin abilities


AnnikaCodes:
Thanks!

PR #496, opened 2022-12-01
Fix ruin abilities

thejetou:
~~They were previously in `Field` but were never initialized. I did not fix them by initializing them because they're not really `Field` properties because unlike terrains it doesn't effect every Pokemon on the field.~~

~~The aura abilities are also in `Field` (though those are actually field properties) and they're never initialized but they work because the abilities are checked in the relevant mechanics files. I could have removed them but maybe they serve a purpose I'm not aware or should serve a purpose later on down the road if there's a Dark/Fairy Aura button in the field control area.~~

### DIFF
--- a/src/index.template.html
+++ b/src/index.template.html
@@ -667,7 +667,7 @@
                     <input class="visually-hidden terrain-trigger calc-trigger" type="checkbox" name="terrain" value="Misty" id="misty" /><label class="btn btn-xxxwide btn-right" for="misty">Misty Terrain</label>
                     <input class="visually-hidden terrain-trigger calc-trigger" type="checkbox" name="terrain" value="Psychic" id="psychic" /><label class="btn btn-xxxwide btn-mid" for="psychic">Psychic Terrain</label>
                 </div>
-                 <div aria-labelledby="selectRuinInstruction" class="gen-specific g9 format-specific doubles" role="group" style="margin: 5px auto auto;" title="Select the ruin abilities from other Pok&eacute;mon on the field.">
+					 <div aria-labelledby="selectRuinInstruction" class="gen-specific g9 format-specific doubles" role="group" style="margin: 5px auto auto;" title="Select the ruin abilities from other Pok&eacute;mon on the field.">
                     <span class="visually-hidden" id="selectRuinInstruction">Select the active Ruin abilities from other Poke&eacute;mon on the field.</span>
                     <input class="visually-hidden calc-trigger" type="checkbox" name="ruin" value="Beads" id="beads" />
 						  <label class="btn btn-xxxwide btn-left" for="beads">Beads of Ruin</label>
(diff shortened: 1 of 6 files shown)


ajhyndman:
Thanks @TheJetOU.  You're right, I added these to field but didn't add an option to the UI to set the values.

Just to explore the options a little: With this approach, can I still run a calc for something like Protosynthesis Flutter Mane with a Chi-Yu ally?  That's what I was worried about when I decided to propose them as field properties instead of just checking the user's ability.

ajhyndman:
Yeah, I understand the UI doesn't support specifying allies.  My point is that making Ruin abilities toggle-able field effects (like dark aura and fairy aura) allows you to calculate the above scenario, while inferring the effect from the ability would not.

I do concede that it could be a little less intuitive not having the ability automatically take effect, but this would be consistent with how the calculator treats dark aura and fairy aura today.

TheJetOU:
I'm a bit confused on what you mean. Why are you not able to calculate a Ruin ability Pokemon using a spread move vs an ally? Ruin abilities affect all Pokemon on the field including allies so I don't see the issue.

>My point is that making Ruin abilities toggle-able field effects (like dark aura and fairy aura)

Dark Aura do not have a toggle-able field effect. They currently work just like the Ruin abilities in my PR - the field effect properties do no seem to do anything.

ajhyndman:
Oh, you're right.  Nobody ever added the field toggle buttons for Dark Aura/Fairy Aura either.  That's probably an oversight?  :shrug:

Sorry, I guess my example wasn't very clear.  The scenario I'm imagining is:

You are trying to calc the damage of Flutter Mane using Moonblast on a target, and you want to account for the fact that Flutter mane both has its own ability (Protosynthesis) and that there is another Pokemon on the field applying the Beads of Ruin effect, reducing the target's special defense.  That's what I couldn't figure out how to calc without a field property.

Maybe the ideal solution is actually to both read abilities and support a field effect?

TheJetOU:
Thanks for the elaboration; I understand the issue now.

My only concern is that adding the two (three?) Aura abilities, four Ruin abilities toggle buttons and others similar abilities (like Steely Spirit and Power Spot) would cause even more bloat to the already cockpit controls looking fields toggles.

I'm less interested in working on that though because I don't play Doubles.

Tagging @AustinXII and @KrisXV for their thoughts on this.

KrisXV:
I definitely think button toggles for these would be beneficial, there are some buttons that shouldn't exist in gen 9 so I'll push a fix for that separately

thejetou:
@KrisXV

src/index.template.html
```diff
@@ -647,7 +647,7 @@
                 <legend align="center">Field</legend>
                 <div aria-labelledby="selectBattleFormatInstruction" class="gen-specific g3 g4 g5 g6 g7 g8 g9" role="radiogroup" style="width: 10.6em; margin: 0 auto 5px;" title="Select the battle format.">
                     <span class="visually-hidden" id="selectBattleFormatInstruction">Select the battle format.</span>
-                    <input class="visually-hidden calc-trigger" type="radio" name="format" value="Singles" id="singles-format" checked="checked" />
```
KrisXV, line 650: Why’d you remove this?

KrisXV: Nvm I read further

PR #497, opened 2022-12-03
Adds Floette-Eternal

ben-tk:
Floette-E is unreleased but is reasonably common enough in draft that I feel adding it to the calculator would be beneficial.

It causes some test fails without the change to gen.ts and I'm not sure if there's a better solution


AnnikaCodes:
Thanks!

PR #498, opened 2022-12-06
Specify "not final evolution" for all Gen 9 pokemon

ajhyndman:
Marks all new and returning pokemon with the nfe attribute.  This is required for eviolite to function correctly.

### DIFF
--- a/calc/src/data/species.ts
+++ b/calc/src/data/species.ts
@@ -8742,12 +8742,10 @@ delete SS['Pikachu-Starter'];
 delete SS['Eevee-Starter'];
 
 const SV_PATCH: {[name: string]: SpeciesData} = extend(true, {}, SS, {
-  // pokemon with new eveolutions
   Bisharp: {nfe: true},
   Dunsparce: {nfe: true},
   Girafarig: {nfe: true},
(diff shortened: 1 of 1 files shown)


calc/src/data/species.ts
```diff
@@ -8742,6 +8742,12 @@ delete SS['Pikachu-Starter'];
 delete SS['Eevee-Starter'];
 
 const SV_PATCH: {[name: string]: SpeciesData} = extend(true, {}, SS, {
+  // pokemon with new eveolutions
+  Bisharp: {nfe: true},
+  Dunsparce: {nfe: true},
+  Girafarig: {nfe: true},
+  Primeape: {nfe: true},
+
   Annihilape: {
```
KrisXV, line 8751:
```suggestion
const SV_PATCH: {[name: string]: SpeciesData} = extend(true, {}, SS, {
  Bisharp: {nfe: true},
  Dunsparce: {nfe: true},
  Girafarig: {nfe: true},
  Primeape: {nfe: true},
  Annihilape: {
```

AnnikaCodes:
thanks

PR #504, opened 2022-12-19
Support Terastallization in the UI

TheJetOU:
- Adds a checkbox for Terastall types
- Allow importing sets with tera types
- Fixes a bug where `Pokemon#hasType` would not return the tera type but the original leading to bugs like Flying Iron Hands taking reduced damage from Outrage in Misty Terrain. I'm not sure if there are any negative ramifications from this change. @DaWoblefet - are there any interactions where the Pokemon's tera type is ignored?

### DIFF
--- a/calc/src/desc.ts
+++ b/calc/src/desc.ts
@@ -14,9 +14,11 @@ export interface RawDesc {
   attackerAbility?: string;
   attackerItem?: string;
   attackerName: string;
+  attackerTera?: string;
   defenderAbility?: string;
   defenderItem?: string;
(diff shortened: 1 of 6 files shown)


TheJetOU:
I cannot seem to reproduce this failing test on my branch locally with `npm run test` so I'm not sure what's going on. I'll check to see if this a failing test within the repo tomorrow by running it on a fresh clone.

KrisXV:
I'm gonna merge this, if Dawob gets back we can always push a fix in the future

PR #506, opened 2022-12-21
Import: Update for Gen 9

thejetou:


monsanto:
Are the version bumps needed for anything? Otherwise it seems OK

thejetou:
When I ran `npm install`, I had some peer deps error for  `eslint-plugin-jest` and `@types/node` was complaining about `AbortError` during compilation. Updating the deps fixed the problems.

I should have probably explained those issues in the commit.

PR #507, opened 2022-12-22
Fix tera type default selection

thejetou:


src/js/shared_controls.js
```diff
@@ -499,6 +498,7 @@ $(".set-selector").change(function () {
 		}
 		if (regSets || randset) {
 			var set = regSets ? correctHiddenPower(setdex[pokemonName][setName]) : randset;
+			pokeObj.find(".teraType").val(set.teraType);
```
thejetou, line 501: On second look, `set.teraType` is going to be `undefined` on non-Gen 9 randsets so I'll look into it tommorow

KrisXV: randsets is different for gen 9, look at how i iterated for moves for gen 9 randsets to get an idea to get teraTypes

thejetou: https://github.com/smogon/damage-calc/pull/507/commits/a03fb2408f8f936a2aa35c9b5cb5eb03118e73f2

PR #508, opened 2022-12-29
Fix tera not keeping original STABs

thejetou:
Regression caused by 151ef66e14ce9295efd04bed3b1f64ae9b59a45c


KrisXV:
T Hank y you

Issue #511, opened 2023-01-04
RBY 0 damage bug


Tree69420:
In RBY, only attacks that are double resisted can do 0 damage. The calculation should have a check between the type effectiveness multiplication and damage range calculation to not do the damage range calculation if the damage is already 1. 

Source: https://bulbapedia.bulbagarden.net/wiki/Damage , and also showdown itself being right (0 dv/statexp level 7 chansey rage always does 1 damage to normal rhydon, rather than usually 0 as the calc claims)

PR #513, opened 2023-01-08
Gens 1-2: If damage is 1, skip random factor

livid-washed:
In Gen 1, the random factor multiplication should be skipped if the damage at that point is 1. In Gen 2, the final damage is rounded up to 1 if it is 0.

This is implemented in PS.

Deals with the issue: https://github.com/smogon/damage-calc/issues/511


AustinXII:
Looks good!

PR #518, opened 2023-01-16
Add type effectiveness precedence for Gen 2, 3, and 4

wartab:
In Gen 2, 3, and 4, when it comes to calculating the damage based on dual type effectiveness, the defensive types are reordered according to a fixed order. This can alter rounding for the computed damage, since type effectiveness multiplications are performed for each type separately.

```
[
    'Normal',
    'Fire',
    'Water',
    'Electric',
    'Grass',
    'Ice',
    'Fighting',
    'Poison',
    'Ground',
    'Flying',
    'Psychic',
    'Bug',
    'Rock',
    'Ghost',
    'Dragon',
    'Dark',
    'Steel',
]
```

This array determines the type order. For example if the defending type is Type 1 = Rock, Type 2 = Ground, the game will swap them for type effectiveness matters.

However, there is an exception when the attacking type is Ice and one of the defending types if Fire. In that case, Fire will always become the second defensive type.

(We have noticed and confirmed this thanks to speedruns of HGSS)

### DIFF
--- a/calc/src/mechanics/gen12.ts
+++ b/calc/src/mechanics/gen12.ts
@@ -66,18 +66,11 @@ export function calculateRBYGSC(
   let secondDefenderType = defender.types[1];
 
   if (secondDefenderType && firstDefenderType !== secondDefenderType && gen.num === 2) {
-    if (move.type === 'Ice' && (firstDefenderType === 'Fire' || secondDefenderType === 'Fire')) {
-      if (firstDefenderType === 'Fire') {
-        firstDefenderType = secondDefenderType;
-        secondDefenderType = 'Fire';
-      }
-    } else {
       const firstTypePrecedence = typeEffectivenessPrecedenceRules.indexOf(firstDefenderType);
       const secondTypePrecedence = typeEffectivenessPrecedenceRules.indexOf(secondDefenderType);
 
       if (firstTypePrecedence > secondTypePrecedence) {
         [firstDefenderType, secondDefenderType] = [secondDefenderType, firstDefenderType];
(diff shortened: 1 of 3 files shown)


DaWoblefet:
I don't know anything about damage calculation prior to 5.

wartab:
image

You can see the damage dealt here is 137, which is not a possible damage output, but if you invert the type on Bronzong, it works.

Proof: https://www.youtube.com/watch?v=-ZOXLocvpAg&t=3969s


I don't know about Gen 1.

For Ice/Fire, that's info I got from a Japanese speedrunner and I'm coming back to you once I know more.

wartab:
Also feel free to not call it precedence, I'm notoriously bad at coming up with names for stuff like that

wartab:
For Ice and Fire interaction, I got this as a source: https://wiki.ポケモン.com/wiki/%E3%83%80%E3%83%A1%E3%83%BC%E3%82%B8#.E7.9B.B8.E6.80.A7.E8.A3.9C.E6.AD.A3

If you don't trust it, someone will probably have to test that interaction.

monsanto:
Interesting find and thanks also to pre for looking further into it. I'm waiting on follow-up on the Ice/Fire question before merging

AustinXII:
@monsanto i was waiting for a response but if there isn’t one tonight I’ll be removing the fire/ice part and will merge

wartab:
I won't have time to investigate this in the near future, feel free to remove it.

AustinXII:
Thanks for this!

PR #519, opened 2023-01-19
Add Adamant Crystal, Lustrous Globe, Griseous Core effects

Zrp200:
Griseous Core and Griseous Orb as coded here work with both formes of Giratina, to ensure backwards compatibility with gens 7,8


Zrp200:
This now passes a run of `npm test`.

monsanto:
Thanks

PR #520, opened 2023-01-20
Implement Ability Shield

Zrp200:
Under this implementation, if the ability is a Mold Breaker ability, it will still be shown even as it is ignored, and the Ability Shield will also be shown.

### DIFF
--- a/calc/src/mechanics/gen789.ts
+++ b/calc/src/mechanics/gen789.ts
@@ -125,7 +125,11 @@ export function calculateSMSSSV(
   if (!defenderIgnoresAbility && !defender.hasAbility('Poison Heal') &&
     (attackerIgnoresAbility || moveIgnoresAbility)) {
     if (attackerIgnoresAbility) desc.attackerAbility = attacker.ability;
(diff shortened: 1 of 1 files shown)


AustinXII:
I think the problem with showing abilities that are being ignored will be users thinking there’s a bug with the calc

urkerab:
(Out of interest I checked current behaviour which is that if you use Mold Breaker against Multiscale then Multiscale isn't shown but if you use it against Shadow Shield then Mold Breaker isn't shown.)

Zrp200:
The way I see it, it might be confusing for the ability to arbitrarily not show up, given Ability Shield is an item and not an ability. All these ability interactions can prove quite confusing if not shown ("why is Turboblaze not ignoring Flash Fire? isn't this a bug?") I'm of the opinion more information is better than less information, as the answer can simply be "Ability Shield".

However, it is true that it does make calculations longer while still giving the same effect in the end. If you'd like me to omit the interaction l, I'll change the code; it's a one line fix.

AustinXII:
sure we can try this

PR #521, opened 2023-01-20
Fix Collision Course/Electro Drift

Zrp200:
Tera type is now used instead of actual type if provided, fixing a bug where Collision Course gave its boost when supereffective against the original type.

### DIFF
--- a/calc/src/mechanics/gen789.ts
+++ b/calc/src/mechanics/gen789.ts
@@ -890,18 +890,24 @@ export function calculateBPModsSMSSSV(
       attacker.hasAbility('Scrappy') || field.defenderSide.isForesight;
     const isRingTarget =
       defender.hasItem('Ring Target') && !defender.hasAbility('Klutz');
-    let effectiveness = 1;
-    for (const type of defender.teraType ? [defender.teraType] : defender.types) {
-      effectiveness *= getMoveEffectiveness(
-        gen,
-        move,
-        type,
-        isGhostRevealed,
-        field.isGravity,
-        isRingTarget
-      );
-    }
-    if (effectiveness >= 2) {
+    const types = defender.teraType ? [defender.teraType] : defender.types;
+    const type1Effectiveness = getMoveEffectiveness(
+      gen,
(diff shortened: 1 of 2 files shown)


Zrp200:
I'm a bit confused why implementing the fix as a for loop broke the code, but it works now.

calc/src/mechanics/gen789.ts
```diff
@@ -890,18 +890,19 @@ export function calculateBPModsSMSSSV(
...
+      types[0],
       isGhostRevealed,
       field.isGravity,
       isRingTarget
     );
-    const type2Effectiveness = defender.types[1] ? getMoveEffectiveness(
+    const type2Effectiveness = types[1] ? getMoveEffectiveness(
       gen,
       move,
-      defender.types[0],
```
Zrp200, line 904: It appears Collision Course is currently using the primary type of the pokemon as both the primary type and the secondary type, which is completely incorrect.

AustinXII:
Did you test this?

Zrp200:
> Did you test this?

I can confirm that the primary type is being used for both types in the case of dual-type Pokémon in the live calculator.

I have also tested the code as implemented in the PR locally and it seems to work as intended.

If you'd like me to add tests ensuring Collision Course/Electro Drift work properly to the repository I can do that.

AustinXII:
> > Did you test this?
> 
> I can confirm that the primary type is being used for both types in the case of dual-type Pokémon in the live calculator.
> 
> I have also tested the code as implemented in the PR locally and it seems to work as intended.
> 
> I'll add some tests in a bit.

Yes some tests would be awesome

Zrp200:
Done. Let me know if the tests need to be adjusted. One thing that proved annoying was that any type changes are lost by `clone`,  otherwise I would have done all the testing on the same Mew. 

The tests check for the presence or absence of a boost in various cases by comparing it to known baseline calculations. Let me know if you'd like me to adjust anything in the tests.

AustinXII:
Thanks!

PR #523, opened 2023-01-22
Add dynamax level support

wartab:
Added no UI for this, can add it later if there's demand for it.


AustinXII:
@KrisXV does this over lap some of the work you were doing?

PR #524, opened 2023-01-23
Implement Big Root Interactions with Draining Moves

### LINKED ISSUE #516: Big Root not working in Generation 9 / SV
Big Root isn't working in SV Damage Calc because HP potentially recovered isn't changing.
image

Zrp200:
This should resolve #516.

I added a test for the draining interactions near the other draining interaction tests. I did not add a test for the leech seed interaction, though, since I wasn't sure where it would go.

### DIFF
--- a/calc/src/desc.ts
+++ b/calc/src/desc.ts
@@ -558,11 +558,13 @@ function getEndOfTurn(
   }
 
   if (field.attackerSide.isSeeded && !attacker.hasAbility('Magic Guard')) {
+    let recovery = Math.floor(attacker.maxHP() / (gen.num >= 2 ? 8 : 16));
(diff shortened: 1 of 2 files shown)


AustinXII:
Doesn’t big root also need some work done in the `mechanics` files ?

Zrp200:
> Doesn’t big root also need some work done in the `mechanics` files ?

I don't see where the changes would be made. Big Root multiplies the recovery from Leech Seed and draining moves by 1.3 after the recovery amount is calculated. It doesn't affect the actual damage of any moves.

AustinXII:
Ah I thought it also increased damage of draining moves

AustinXII:
Thanks!

PR #528, opened 2023-02-24
Show (and use) EVs in the calculator for Gen 1/2

Sabelette:
As per my post on Smogon: 
"Small thing but could we please have EVs shown on calc for gens 1-2? This is actually relevant for some calcs - for example in gen 1 "Safetwo" (immune to overflow) uses 341 special, not 406, which is literally impossible to set in the calc without Stat EXP/EVs, and a lot of Counter techs require lower stats than just dropping the DV to 1. Editing base stats is often a little inaccurate and is super inconvenient since it resets every time you change mons. Just showing it and defaulting all sets to max in all stats would help a lot."

I attempted to: 1. Remove the automatic +63 replacing EVs for gen1/2 and use the same EV function as gen3+, 2. make the calc check EV value for gen1/2 rather than sub in 252, 3. make the EV UI show in gen1/2, and 4. default EVs to 252 for blank imports rather than 0 like later gens since most Pokemon will use max stats in gen 1/2. I am especially unsure I did that last part right.

I am not a coder and cannot promise this is even a functional fix but I tried to read through the files and ensure I didn't miss anything. I also have never used GitHub so sorry if I did something wrong with the updates/pull request, I'm playing this by ear. Hopefully it's either semi-functional or a springboard for someone else to add EVs to the calc in a better way. Thanks for reading!

### DIFF
--- a/calc/src/stats.ts
+++ b/calc/src/stats.ts
@@ -159,9 +159,9 @@ export const Stats = new (class {
 
   calcStatRBYFromDV(stat: StatID, base: number, dv: number, level: number) {
     if (stat === 'hp') {
-      return Math.floor((((base + dv) * 2 + 63) * level) / 100) + level + 10;
+      return Math.floor((((base + dv) * 2 + Math.floor(ev / 4)) * level) / 100) + level + 10;
     } else {
-      return Math.floor((((base + dv) * 2 + 63) * level) / 100) + 5;
+      return Math.floor((((base + dv) * 2 + Math.floor(ev / 4)) * level) / 100) + 5;
     }
   }
 
--- a/src/honkalculate.template.html
+++ b/src/honkalculate.template.html
@@ -117,7 +117,7 @@
                                 <th></th>
                                 <th>Base</th>
                                 <th class="gen-specific g3 g4 g5 g6 g7 g8 g9">IVs</th>
-                                <th class="gen-specific g3 g4 g5 g6 g7 g8 g9">EVs</th>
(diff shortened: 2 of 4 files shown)


AustinXII:
Did you test this by chance?

wartab:
It uses a variable `ev` that's not declared, so it can't work. 
Also, the actual implementation uses Stat Exp instead of EVs, even if EVs would be the square root of Stat Exp eventually. If such a feature existed, it would be useful to actually use Stat Exp. Stat Exp is a value reaching from 0 to 65535 (the 63 comes from sqrt(65535) / 4 btw).

Sabelette:
Frankly I have no clue how to test this effectively as again, I’m not really a programmer; as for the choice to use EVs yes I’m aware that the older games use Stat Exp but since the actual implementation on Pokémon Showdown uses EVs just like later gens and there is no point to asking users to input their exact stat exp amount, it makes more sense to me to just use EVs and streamline importing.

I was really just hoping that this would draw some attention to the issue as it makes calcing really rough for some old gen interactions and could be fixed by just letting us change the EV/Stat Exp value like every other gen can, but it would take someone more competent to actually implement it properly.

Sabelette:
> It uses a variable `ev` that's not declared, so it can't work.

I have hopefully fixed this part now; I intend to learn more so I can make effective contributions but in the meantime I hope this is functional while I learn the best way to test my changes.

AustinXII:
could you write a test for this?

PR #532, opened 2023-03-12
[@smogon/calc] Speed up npm test script

ajhyndman:
I suspect this wasn't always the case, but `npm test` no longer depends on compiled artifacts.

Since that's the case, there's no need to run the compiler before every test run.  Removing this build step speeds up the development + testing loop significantly!

### DIFF
--- a/calc/package.json
+++ b/calc/package.json
@@ -34,7 +34,6 @@
     "test": "jest",
     "coverage": "jest --coverage",
     "prepare": "npm run build",
-    "pretest": "npm run build",
     "posttest": "npm run lint"
   }
 }


shrianshChari:
I figured that I would check this out, as to whether or not removing the pretest step would speed up the testing. Spoiler alert: it does.

**Methodology:**
1. Delete `node_modules` folder
2. `npm i`
3. Run `time npm test`
4. Record the "total" time of execution.
5. Repeat this, but delete the `pretest` script before running the test.

**Results:**
I did 5 consecutive tests both with and without the pretest step. With the pretest step, the 5 tests ran at a mean of 44.436 seconds, while without it, the tests ran at a mean of 21.501 seconds. None of the tests failed, meaning that removing the pretest step didn't change the behavior of these tests. 

<details>
  <summary>Screenshots</summary>
  
Screenshot from 2023-10-07 13-33-02
Screenshot from 2023-10-07 13-33-13
</details>

PR #535, opened 2023-03-14
Improve Weight Handling

Zrp200:
* Moved all weight modification code to one place (utils)
* Heavy/Light Metal and Float Stone can now stack.
* Heavy Metal, Light Metal, and Float are now shown in the description when applicable 
* Weight modification is now done in Hectograms, though it still returns Kilograms after truncating fractional hectograms (Showdown implementation uses truncation, not rounding). 
* It is now impossible to get a weight lower than 0.1kg.

I also added tests for weight handling that mirror Showdown's.

This should address the weight-based inaccuracies expressed in #443


Zrp200:
Bump

Zrp200:
Rebased to fix the merge conflict. It was a trivial import conflict.

PR #537, opened 2023-03-17
Added option to manually specify Quark Drive / Protosythesis stat boost

Zrp200:


calc/src/test/calc.test.ts
```diff
@@ -956,6 +956,21 @@ describe('calc', () => {
           defender.teraType = 'Normal';
           expect(calc(cc)).toEqual(se);
         });
+        test('Quark Drive should not take into account boosted stats', () => {
```
AustinXII, line 959: We should be testing both

Karthik99999:
I'm like 99% sure that these abilities are actually supposed to take stat boosts into account when determining what stat they boost. Where did you find that they didn't take stat boosts into account?

Zrp200:
> I'm like 99% sure that these abilities are actually supposed to take stat boosts into account when determining what stat they boost. Where did you find that they didn't take stat boosts into account?
@Karthik99999 

This was done in response to [this forum post on the Damage Calc thread](https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/post-9544846)

I see that Bulbapedia has it listed as taking into account stat stages, though:
> When a Pokémon with Quark Drive is on the field and [Electric Terrain](https://bulbapedia.bulbagarden.net/wiki/Electric_Terrain_(move)) is active, or the Pokémon is holding [Booster Energy](https://bulbapedia.bulbagarden.net/wiki/Booster_Energy), its highest [stat](https://bulbapedia.bulbagarden.net/wiki/Stat) (other than [HP](https://bulbapedia.bulbagarden.net/wiki/HP) and taking into account [stat stages](https://bulbapedia.bulbagarden.net/wiki/Stat_modifier), but not [held items](https://bulbapedia.bulbagarden.net/wiki/Held_item)) is increased by 30%, or 50% if [Speed](https://bulbapedia.bulbagarden.net/wiki/Stat#Speed) is the highest stat. Upon activation, a message tells which stat is increased to all the players.
https://bulbapedia.bulbagarden.net/wiki/Quark_Drive_(Ability)

Does this mean that Showdown has it implemented incorrectly? Can it actually change?

calc/src/test/calc.test.ts
Zrp200: Both quark and Proto? In that case I'll change the implementation from Booster Energy to the respective field effects, which should make the tests more powerful.

Karthik99999:
The actual issue here is that the calc doesn't know when exactly the boost has been applied. so when you change the boosts on the calc, it modifies the boosted Protosynthesis/Quark Drive stat accordingly, when in actuality the boost cannot change once it has been set (unless the ability reactivates). Showdown has the correct behavior, which is to boost based off raw stat + stat boosts. There should probably be a way on the UI to override what stat is being boosted on the calc, rather than changing this behavior of picking which stat to automatically boost with the ability (which again, is correctly taking stat boosts into account).

Zrp200:
Okay, this should now be entirely correct. Instead of ignoring stat changes, this now has the existing (correct) behavior by default, but allows the stat to also be manually specified via a dropdown menu (much like Supreme Overlord).

thejetou:
Can you add the dropdown to honkcalculate too?

thejetou:
Thank you!

PR #542, opened 2023-04-07
Improvements to multi-hit moves (motivated by Population Bomb) 

lsparlin:
I discovered that the gen 9 move "Population Bomb" is a Muti-hit move, but the damage calculator was only able to display the 10-hit damage result.  Digging deeper I discovered the reason for this was the javascript was only able to handle 2-5 hit moves. I've added a few changes to make move hit more robust (could technically support 3-8 hits in the future) and added a range instead of a static `10` to "Population Bomb"

# Updates
- Change Population bomb to use a range style hits property.
  - `[2, 10]` instead of just `10`
- improve logic for showing number of hits in calc

## Icicle Spear (2-5 hits still works)
Screenshot 2023-04-07 at 3 31 13 PM

## Population Bomb (2-10 hits now supported)
Screenshot 2023-04-07 at 3 30 57 PM

### DIFF
--- a/calc/src/data/moves.ts
+++ b/calc/src/data/moves.ts
@@ -4528,7 +4528,7 @@ const SV_PATCH: {[name: string]: DeepPartial<MoveData>} = {
     maxPower: 90,
     makesContact: true,
     isSlicing: true,
-    multihit: 10,
+    multihit: [2, 10],
   },
   Pounce: {
     bp: 50,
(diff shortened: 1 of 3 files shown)


Karthik99999:
#458 already does this

lsparlin:
> #458 already does this

Ah thanks for letting me know

PR #546, opened 2023-05-01
Protosynthesis / Quark Drive Auto-Select & Override

Ericzklm:
Fixed typo in Protosynthesis that prevented the dropdown from being shown


Ericzklm:
I noticed that overrides in Protosynthesis / Quark Drive are still only active if the weather / terrain is in play or if booster energy is held.

```
if ((attacker.hasAbility('Protosynthesis') &&
        (field.hasWeather('Sun') || attacker.hasItem('Booster Energy'))) ||
        (attacker.hasAbility('Quark Drive') &&
            (field.hasTerrain('Electric') || attacker.hasItem('Booster Energy')))) {
        if ((move.category === 'Physical' &&
            (0, util_2.getQPBoostedStat)(attacker) === 'atk') ||
            (move.category === 'Special' && (0, util_2.getQPBoostedStat)(attacker) === 'spa')) {
            atMods.push(5325);
        }
    }
```

Should the override ignore the weather / terrain and booster energy if one of the options is selected? For example, if Flutter Mane enters a weatherless field with booster energy, the booster energy is consumed, and symbiosis passes over a different item.

thejetou:
Oh, I'm so sorry, I fixed [this](https://github.com/smogon/damage-calc/commit/26fe67b7a3ff46c5aa9b2b333b0c5db72b59f85f) just now without checking if there were any open PRs first (I should turn on notifications for this repo...).

>Should the override ignore the weather / terrain and booster energy if one of the options is selected? For example, if Flutter Mane enters a weatherless field with booster energy, the booster energy is consumed, and symbiosis passes over a different item.

That is a good point. I think to tackle this issue, you would need to add something to the effect of "Inactive" on the dropdown menu and make it automatically "Inactive" if there's no Booster Energy / Electric Terrain / Sun and "Auto-Select" if in the presence of the aforementioned conditions. Because, if you currently make the override always activate the ability, then all Quark Drive / Protosynthesis Pokemon will have a boosted stat with no way to turn it off other than changing the ability.

Thank you for your interest in the calc and I hope this mishap doesn't discourage you!

Ericzklm:
Made changes to describe the following behavior:

- If Protosynthesis / Quark Drive is selected without the necessary conditions to activate, the status is _inactive_
  - If the item is changed to booster energy or the correct weather / terrain is set, the status becomes _auto select_
  - If the item / weather / terrain are removed, the status reverts to _inactive_
  - If a specific stat is selected as the override, that stat is increased by the appropriate multiplier
    - Overriden selections are not affected by changing item / weather / terrain

Added Protosynthesis / Quark Drive to other calculators
Magic Room / Booster Energy interaction is not implemented

thejetou:
Thank you!

PR #547, opened 2023-05-08
Automatic detect dark mode setting. 

mlyoung25:
Simple accessibility feature to utilize dark mode features on page load.


thejetou:
Thanks!

PR #548, opened 2023-05-16
Restyled checkboxes and buttons in dark mode

shrianshChari:
Now darkens buttons and checkboxes when Showdown is in dark mode.

Checkboxes are unchecked:
Screenshot from 2023-05-16 16-24-14

Checkboxes are checked:
Screenshot from 2023-05-16 16-24-41

In both instances, the unstyled buttons ("Clear Imported Sets", "Export", "Click for  Light Theme") are now visibly darker.

### DIFF
--- a/src/css/dark-theme.css
+++ b/src/css/dark-theme.css
@@ -15,6 +15,7 @@
 body {
     color: var(--text);
     background-color: var(--background);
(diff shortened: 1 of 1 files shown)


thejetou:
Thanks!

PR #549, opened 2023-05-30
Force word wrap on selected sets

vivapuffs:
Fixes problem where selected set names that were too long caused that portion of the calc to be pushed off the page
Before:
image
After:
image


monsanto:
Thanks!

PR #550, opened 2023-06-02
Ruin Pokemon are immune to their own Ruin effects

alotofnekos:
It works correctly in PS, but isnt implemented in the damage calc. See this post https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/post-9625220

This is my first time trying this, thankie in advance.


thejetou:
Thanks!

PR #553, opened 2023-06-05
Fixes to Honkalculate UI to match 1v1 UI

AWildNoob:
These changes serve to make the Gen 9 Honkalculate UI match the 1v1 UI.

The specific changes are:
-Added Tera selector
-Added Stealth Rock button, removed Steelsurge
-Changed Hail to Snow in weather selector in Gen 9
-Added Magic Room/Wonder Room to Honkalculate (from Gens 5 - 9)

I've tested and it seems that simply copying the UI elements from the 1v1 UI seems to work without additional changes.


thejetou:
Thanks!

Issue #555, opened 2023-06-06
Fix Kleavor's base stats


shrianshChari:
https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/post-9650084
https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/post-9662209
In the calculator, it displays Kleavor's base Attack as 130 and base Special Defense as 75, but on Bulbapedia, Serebii, and PS they are 135 and 70, respectively.

PR #557, opened 2023-06-17
Add Hemogoblin & fix Kleavor's stats

shrianshChari:
Doesn't pass `npm test`, would like some advice on letting the test pass since I couldn't figure out how the test works.

Calculator will still work w/ updated stats

Addresses #555 

Sources:
https://serebii.net/pokedex-sv/kleavor/
https://bulbapedia.bulbagarden.net/wiki/Kleavor_(Pok%C3%A9mon)#Base_stats
image

### DIFF
--- a/calc/package-lock.json
+++ b/calc/package-lock.json
@@ -12,7 +12,7 @@
         "@types/node": "^18.14.2"
       },
       "devDependencies": {
-        "@pkmn/dex": "^0.7.25",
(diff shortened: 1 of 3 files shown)


thejetou:
The build was failing because `@pkmn/dex` which the test suite uses to make sure the calc data is correct had an old incorrect version. I updated the pkg and added Hemogoblin to keep the tests passing.

Thanks!

PR #563, opened 2023-07-19
Makes entire background dark in dark theme

shrianshChari:
Currently, when looking at the calculator in its dark theme, the bottom of the calculator (beyond the button to switch the theme from light to dark) would just be white. This somewhat defeats the purpose of having a dark theme if there will still be a large patch of white at the bottom of the page (which is much larger on mobile devices). This fixes this.

Before:
Screenshot from 2023-07-19 10-11-44

After:
image

Screenshots were taken on 1440p monitor and resizing the browser zoom to 80%.

### DIFF
--- a/src/css/dark-theme.css
+++ b/src/css/dark-theme.css
@@ -12,7 +12,7 @@
     --column-1: #2a2a2a;
 }
 
-body {
+html, body {
     color: var(--text);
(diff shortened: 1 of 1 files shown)


monsanto:
Seems OK to me. Thanks

PR #568, opened 2023-07-27
Change Revelation Dance's type when attacker has Terastallized

shrianshChari:
Fixes https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/post-9716660

Resulting calc: 0 SpA Tera Water Oricorio-Pom-Pom Revelation Dance vs. 252 HP / 0 SpD Sandaconda: 254-302 (72.9 - 86.7%) -- guaranteed 2HKO after Leftovers recovery

Can create tests if needed

### DIFF
--- a/calc/src/mechanics/gen789.ts
+++ b/calc/src/mechanics/gen789.ts
@@ -173,7 +173,11 @@ export function calculateSMSSSV(
     desc.terrain = field.terrain;
     desc.moveType = type;
   } else if (move.named('Revelation Dance')) {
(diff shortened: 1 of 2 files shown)


AnnikaCodes (approved):
This looks good to me. If you can write tests, that would be great — but if not, this can be merged without them.

shrianshChari:
I'll try and write some since testing code is always good.

AnnikaCodes:
That's excellent!

PR #569, opened 2023-08-01
Remove Enigma Berry from GSC only list

### LINKED ISSUE #472: Duplicate Items in Gen 8
https://github.com/smogon/damage-calc/blob/bdf9e8c39fec7670ed0ce64e1fb58d1a4dc83b73/calc/src/data/items.ts#L470-L471

This is adding in Enigma Berry twice...

Think it can be fixed by just removing this line https://github.com/smogon/damage-calc/blob/bdf9e8c39fec7670ed0ce64e1fb58d1a4dc83b73/calc/src/data/items.ts#L87 since I don't think Enigma Berry even exists in gen 2?

Found when I was writing an API wrapper using AJV

Logs: 

```Error: schema is invalid: data/oneOf/7/properties/item/enum must NOT have duplicate items (items ## 66 and 513 are identical)```
```513 Enigma Berry```
```66 Enigma Berry```

shrianshChari:
Fixes #472 

Will make sure the Enigma Berry doesn't appear twice in Gens 8 and 9.

### DIFF
--- a/calc/src/data/items.ts
+++ b/calc/src/data/items.ts
@@ -83,7 +83,6 @@ const GSC_ONLY = [
   'Pink Bow',
   'Polkadot Bow',
   'PSN Cure Berry',
-  'Enigma Berry',
 ];
 
 const ADV = GSC.filter(i => !GSC_ONLY.includes(i)).concat([


lawauditswe (approved):

thejetou:
Thanks!

PR #570, opened 2023-08-01
Remove duplicate spaces in the calc description

shrianshChari:
https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/post-9724268

Before (first line) vs after (second line) with double spaces highlighted.
image

### DIFF
--- a/calc/src/desc.ts
+++ b/calc/src/desc.ts
@@ -953,6 +953,7 @@ function buildDescription(description: RawDesc, attacker: Pokemon, defender: Pok
   if (description.isWonderRoom) {
     output += ' in Wonder Room';
   }
+  output = output.replace(/[ ]{2}/g, ' ');
   return output;
 }
 


AnnikaCodes:
Shouldn't we be avoiding the double spaces in the first place rather than using a regex to remove them afterwards?

shrianshChari:
That's probably a better solution, but I didn't want to touch as much of the existing code when I did this commit. I can look at it again and figure out something that would avoid double spaces while constructing the calc description.

shrianshChari:
Found that the issue behind the double space problem was that the descriptors for the Flower Gift boosts, Battery boosts, Power Spot boosts, and the switch boost (for Pursuit) had spaces on either side of the message when it should have only been on one side. This is now standardized with the other descriptors, and a test has been written for this.

lawauditswe (approved):

thejetou:
Thanks!

PR #571, opened 2023-08-02
Struggle should be unaffected by Normalize

shrianshChari:
https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/post-9725369


AnnikaCodes:
Thanks!

PR #573, opened 2023-08-08
Implement default level selection

DimK19:
This is a feature requested by issue https://github.com/smogon/damage-calc/issues/348, under "Approved Features". The purpose is to have a quick and permanent way of switching between the two most commonly used level numbers, 100 and 50. I placed the radio buttons for this selection in the central column:

Screenshot
image

It may not be exactly a property of the field, but this position indicates that it applies to both parties. Additionally, I have made the radio group gen-specific to all generations from 4 and onwards, since before that there were not any game formats played at level 50 as far as I am aware.

The default value when the page loads is 100. Upon changing it, the two input fields of the levels are updated and the stats recalculated accordingly. When changing to another set from the drop-down list, priority is given to the predetermined level of that set, meaning 100 for ou, uu, etc, 50 for battle spot etc, 5 for lc, and so on, and not the default level indicated by the radio button. This is done so as to not interfere and tamper with the saved sets. When a "Blank Set" is selected, then its level will be the chosen default level. Your input on this last decision would be appreciated, because even though I think this is the most logical approach, it may confuse some users.

Please let me know if any changes must be made. Thanks.


shrianshChari:
I just have a couple of suggestions. First, I'd include an option for level 5 for Little Cup builders. Second, make it so that when someone hovers over the button, it'll say "Blank sets will be level 100/50/5" so that people are aware that only the blank sets get affected.

DimK19:
> make it so that when someone hovers over the button, it'll say "Blank sets will be level 100/50/5" so that people are aware that only the blank sets get affected.

@shrianshChari In the screenshot I have attached, if I were to click Level 50, the levels would instantly change to 50, even though the selected sets are not blank. The limitation on blank sets is applied when changing a set from the list. Perhaps it would be simpler to impose the selected level universally on all sets, overriding the level of the saved sets. I suppose that should be decided by the maintainers.

> include an option for level 5 for Little Cup builders

Certainly I could add this, are there any other levels that should be included?

shrianshChari:
> Perhaps it would be simpler to impose the selected level universally on all sets, overriding the level of the saved sets. I suppose that should be decided by the maintainers.
I do like this since it's simpler; if you do so then create a selection that just leaves the levels as they are for the sets.

> Are there any other levels that should be included?
I think most metagames just follow levels 100, 50, and 5. If we need more levels in the future, then we can just add them.

thejetou:
I'd prefer for saved sets to override the default selection. Otherwise, if Level 100 is automatically picked then VGC/BSS sets wouldn't work. I suppose you could have a "None" button, but generally I think saved sets should override calc settings.

src/js/shared_controls.js
```diff
@@ -100,8 +100,17 @@ $("input:radio[name='format']").change(function () {
...
+	defaultLevel = $("input:radio[name='defaultLevel']:checked").val();
+	$("#levelL1").val(defaultLevel);
+	$("#levelR1").val(defaultLevel);
+	$("#levelL1").trigger("change");
+	$("#levelR1").trigger("change");
+});
+
 // auto-calc stats and current HP on change
-$(".level").keyup(function () {
+$(".level").bind("keyup change", function () {
```
thejetou, line 113: Why? What's the difference?

DimK19:
@thejetou As it is now, saved sets will override the default level selection when initially chosen in the calculator (as per your preference) but if the user clicks a different default level **afterwards**, the level of both sets will change accordingly. For example if I have set the default level to 100 and choose from the drop-down list a saved set with level 99, then it will have level 99 (ignoring the default), but if then I click level 50, both active sets will have their levels changed to 50, regardless of whether they are saved or not. I think this approach provides ease of use (not having to scroll down to the blank set) without excessively tampering with the saved sets.

Practically the user who is interested in a level-100 format may never need the functionality (as sets are level 100 by default unless otherwise specified – in which case their level is preserved), and the user who is interested in a level-50 format can choose any two sets from the list and then bring them down to level 50 with a single click.

src/js/shared_controls.js
DimK19: I can explain this, the purpose is to have both active sets’ stats change automatically when the default level is changed, in accordance with that level. The `keyup` event is not triggered when the level values are changed programmatically, hence the need of the addition of the `change` event, so that the stats and the damage output are recalculated.

DimK19:
@thejetou  After @shrianshChari 's suggestion, I have added a third option for level 5. Apparently there are level 5 formats in all generations. Below are two screenshots:

image
_Generations 1-3_
image
_Generations 4 and on_

As I have mentioned, I have made the option for level 50 generation-specific to the 4th and onwards, as I did not find any formats prior to that played at level 50. Of course I could change that specification if necessary.

Thanks.

src/js/shared_controls.js
thejetou: Oh I see, that makes sense.

thejetou:
This should be ready for merge once the merge conflicts are fixed.

shrianshChari:
Personally, I still feel having a "None" or "Default" button is good, since it gives the user a visual that the level selector is not active. Then, you would just have the level selector default to "None" and the level would override every set. However, I also think that we could do some user testing (if we wanted to do that) since we won't know which UX is better without testing. Ultimately, jetou has the final say on the merge.

thejetou:
> Personally, I still feel having a "None" or "Default" button is good, since it gives the user a visual that the level selector is not active. Then, you would just have the level selector default to "None" and the level would override every set. However, I also think that we could do some user testing (if we wanted to do that) since we won't know which UX is better without testing. Ultimately, jetou has the final say on the merge.

The thing is that a level selector is always active (level 100), so having None and Level 100 is kind of redundant IMO. Other than the rare case where you have a Level 60 Pokemon then you click the level 50 selector and would like to back to Level 60 but I'm not confident that this case warrants a button - if it's desired by users later down the line I'll add it in but it doesn't seem worth it for now in my perspective.

DimK19:
> This should be ready for merge once the merge conflicts are fixed.

The merge conflict should be resolved now.

thejetou:
Great work. Thank you!

PR #574, opened 2023-08-17
Blank Sets have Default Abilities

FomTarro:
I have to assume this was a typo; blank sets will now have the default ability for the species/forme instead of `(other)`.


thejetou:
Thanks!

PR #576, opened 2023-08-23
Support Super Fang & Ruination

ForwardFeed:
And i've paid attention to round down if the HP are odd


ForwardFeed:
I'll work on it

ForwardFeed:
Also two question, why aren't the genders fully implemented, i mean is there a specific reason for that? Also i plan to integrate Psywave & Magnitude, but i have no clue how to format the data, should i just show "Lowest BP Roll - Median BP roll - highest BP Roll ", or do like with multi-hits move where there is a menu on the BP rolls and the damage shows normally from there? thanks

calc/src/mechanics/gen12.ts
lawauditswe, line 1: I may be missing something but if `defender.curHP() === 1`, doesn't that mean `lostHP === 0`? 

Super Fang should always do 1 damage at least, and will knock out a Pokemon with 1 HP.

calc/src/mechanics/gen12.ts
```diff
@@ -74,7 +74,13 @@ export function calculateRBYGSC(
     }
   }
 
-
+  //in gen 1, Super fang can hit ghosts
```
thejetou, line 77:
```suggestion
  // In Gen 1, Super Fang can hit Ghost types
```

calc/src/mechanics/gen12.ts
```diff
@@ -74,7 +74,13 @@ export function calculateRBYGSC(
...
+  if (gen.num === 1 && (move.named('Super Fang'))) {
+      const lostHP = Math.floor(defender.curHP() / 2) || 1;
+      result.damage = lostHP;
+      return result;
+  }
```
thejetou, line 82: In Gen 2, you should check if the opposing Pokemon can be hit by Super Fang (probably by `getMoveEffectiveness`) and run the calculation if so.

And also check if the opposing Pokemon has Protect up just like it's done in `gen789.ts`.

calc/src/mechanics/gen3.ts
```diff
@@ -122,6 +122,12 @@ export function calculateADV(
 
   desc.HPEVs = `${defender.evs.hp} HP`;
 
+  if (move.named('Super Fang')) {
```
thejetou, line 125: Same here and in all other gens.

calc/src/mechanics/gen12.ts
thejetou: Yes, but`Math.floor(1 / 2) || 1` equals to `0 || 1` which evaluates to  `1`.

thejetou:
> Also two question, why aren't the genders fully implemented, i mean is there a specific reason for that? Also i plan to integrate Psywave & Magnitude, but i have no clue how to format the data, should i just show "Lowest BP Roll - Median BP roll - highest BP Roll ", or do like with multi-hits move where there is a menu on the BP rolls and the damage shows normally from there? thanks

I'm not sure why the genders aren't full implement honestly, lol. About Psywave & Magnitude, probably just format it "Lowest roll - Highest roll" and include each possible roll? I think that's sufficent.

calc/src/mechanics/gen12.ts
ForwardFeed: That was my bad, i put the calculation of Super fang before the verification of type effectiveness

ForwardFeed: In gen 2 the check for Super fang pass after the check of type effectiveness.

ForwardFeed: The same goes for the field.defenderSide.isProtected check, so as far i can tell, there is no issue with that.

ForwardFeed:
Also for the gender, i was thinking more about the [addSets](https://github.com/smogon/damage-calc/blob/00985cb56cf7846fd42376660d924ae92e516bfe/src/js/moveset_import.js#L273) and [ExportPokemon](https://github.com/smogon/damage-calc/blob/00985cb56cf7846fd42376660d924ae92e516bfe/src/js/moveset_import.js#L12)  in the front end. Rivalry and gender in the UI is well implemented, my bad i missexplained.

PR #578, opened 2023-09-15
Update to Teal Mask Smogon calc support, and fix package bundler

pikalytics:


pikalytics:
Whoops, this was supposed to go to our public fork 😅

PR #441, opened 2021-04-09
Add click-to-copy feature

DimK19:
Made a very small addition using the Clipboard API. I ran `npm test` locally, apparently the two failing tests are unrelated to it. Hope it works properly!

PR #579, opened 2023-09-16
Improve appearance of click-to-copy confirmation message

DimK19:
Since originally authoring the pull request that implemented this feature (https://github.com/smogon/damage-calc/pull/441), I have noticed a few ways it could be better.

**Firstly**, there is an issue when the damage calculation message has a certain length, such that the hidden element that appears on click is pushed to a second line, while no text is there. This results in the abnormality shown in the screenshots below: 

Picture1

Picture2

The layout is pushed down by a blank line that accommodates the hidden element.

In order to fix it, I have changed the position of the confirmation message to appear above the text. I have set its position to absolute, so it will not interfere with the rest of the layout.

Below are three screenshots of how it looks, including two extreme cases of a very short output message, and a very long individual move output. Please let me know if there are any other cases I should consider for the position of the message.

Picture3

Picture4

Picture5

**Secondly**, I reduced the duration of the appearance from 2s to 1.5s, because I found it lingered for too long.

Thanks.

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -180,21 +180,33 @@ th {
     font-weight: bold;
     font-size: 1.1em;
     margin: 0.5em 0;
+    position: relative;
 }
 .big-text #mainResult:hover {  /* indicates the presence of click-to-copy function */
     cursor: pointer;
     opacity: 0.7;
 }
 #tooltipText {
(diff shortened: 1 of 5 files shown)


thejetou:
Thanks!

PR #580, opened 2023-10-03
Fix "No matches found" in dark mode

shrianshChari:
Currently, when you search through the dropdown, when you search for something that has no matches, it should display that there are "No matches found." However, in dark mode, the text is changed to a light color, but the background still displays as a dark background, meaning that you see white text on a white background. This fixes this issue.

Screenshot:
image
image


thejetou:
Thanks!

PR #581, opened 2023-10-07
[@smogon/calc] Fix npm tests

shrianshChari:
Updated @pkmn/dex so that tests work.
Basically test now supports SV DLC 1

### DIFF
--- a/calc/package-lock.json
+++ b/calc/package-lock.json
@@ -1,18 +1,18 @@
 {
   "name": "@smogon/calc",
-  "version": "0.7.0",
+  "version": "0.8.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@smogon/calc",
-      "version": "0.7.0",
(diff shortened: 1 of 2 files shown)


calc/package-lock.json
```diff
@@ -1,18 +1,18 @@
 {
   "name": "@smogon/calc",
-  "version": "0.7.0",
+  "version": "0.8.0",
```
shrianshChari, line 3: I don't know why the version in the package-lock.json changes, I just edited the `package.json`, then did `rm -rf node_modules` and `npm i` and the version modified automatically.

monsanto: Maybe someone forgot to commit a package-lock change? This happened to me too

shrianshChari:
Maybe. I'll try and revert the change on my end and run the tests again.

shrianshChari:
Ok, it seems that my `node_modules` were out of date and the pre-DLC version of @pkmn/dex was installed.

Issue #572, opened 2023-08-03
Feat: Calculate damage for Pain Split


jmynes:
Is it possible to add Pain Split calculation?
It would be great to quickly see if I will heal or deal damage in any particular battle state, without having to manually do the math.

I'm not sure how this well fits in the official calculator's UI, but if damage healed/dealt were queryable, it would be easier to add to other software which queries this package.

----

If it were to fit into the existing UI, I would guess the best option is to put healing % on the left of the `-`, and dmg on the right, maybe with color coding, or a dotted underline indicating a tooltip?

ForwardFeed:
Winning HP:
![image](https://github.com/smogon/damage-calc/assets/76596109/0dab4e0d-8e7b-47f2-bdf1-8cfae3b4af94)
Losing HP:
![image](https://github.com/smogon/damage-calc/assets/76596109/6f3124f6-f3c2-4325-94e0-67f4accbe0ee)

I managed to render this kind of results, if people feels its right so, i'll do a PR

AustinXII:
#585

PR #585, opened 2023-10-29
Support Pain Split

ForwardFeed:
to implement the requested feat #572

Note, i tried to make a full rebase before pushing but failed (i really struggle with git) so there is the other changed of ruination and super fang along side it


jmynes:
This would be super useful in showdex if merged in calc 👀

calc/src/mechanics/gen789.ts
ForwardFeed, line 1: Nature's madness at least does 1hp in damage, it's tiny but the || 1 was justified there

ForwardFeed:
can't you squash it yourself? like, you're the last one that have worked on it makes more sense for me?

Issue #499, opened 2022-12-06
Allow Calcs to be Shared


corydissinger:
It's probably a good bit of work to allow calcs to be shared - but would it be possible using query params / path variables? Effectively a long URL with necessary state so people can share calcs.

chessai:
Would love this.

joebangara43903:
![image](https://user-images.githubusercontent.com/131228993/232965015-897c692b-60a6-4439-a186-7cde90ecad34.png)

ForwardFeed:
Maybe using the import/export field is better,
Just adding a export calculation button and using the import field to just paste a Base64 or JSON Text

ForwardFeed:
Well, i'm doing it, and nevermind i agree the path variable is the most practical for end user

AustinXII:
Duplicate #586

PR #587, opened 2023-11-05
Support Sharing calculation

ForwardFeed:
in reference to #499
I haven't fully tested it however as I don't know too much how should I test it.
in todofix there is this bug with Dynamax and non-full HP situation.


AustinXII:
@thejetou opinions? I'm sure a lot of people would like this feature.

AustinXII:
After some testing, it does seem to works for abomasnow but fails to work if you try sharing a calc with a different pokemon. I'll see if I can figure out whats going on and hopefully we can merge this.

AustinXII:
Oh I see, it only doesn't work if the Pokémon has more than one form

AustinXII:
Also the set should only be shown in the same generation the damage calc was made in. Currently if you shared a damage calc and the person uses the link, that set would be shown in all generations.

ForwardFeed:
When I did the code a while back ago it was supposed to redirect the receiver gen to the sender gen. It is likely not up to date of the current codebase on many points.

AustinXII:
@ForwardFeed I would love to add this if you’d take a jab at updating this PR now that I’m back.

ForwardFeed:
I'm sorry I'll probably not work on the smogon calc in the future. I've made this PR 9 months prio, In between I have made my own fork where i changed a lot of code that i don't remember well the codebase of the smogon calc and currently If  I'm again working on a calc I would want to make my own nuzlocke specialized calc.

AustinXII:
@ForwardFeed no problem at all. We appreciate the PR, I’m sorry for the ghosting and no response.

PR #592, opened 2023-12-26
Gens 4-9: Remove interaction between Natural Gift and Berry Juice

shrianshChari:
Currently, when a Pokemon uses Natural Gift while holding a Berry Juice, it is treated as a 1 BP Normal-type move. This will now accurately display that a Pokemon using Natural Gift while holding a Berry Juice will deal 0 damage.

Abomasnow with Berry Juice using Natural Gift


shrianshChari:
Closing for now to refactor code and use a different branch

PR #593, opened 2023-12-26
Gens 4-9: Remove interaction between Natural Gift and Berry Juice

shrianshChari:
Currently, when a Pokemon uses Natural Gift while holding a Berry Juice, it is treated as a 1 BP Normal-type move. This will now accurately display that a Pokemon using Natural Gift while holding a Berry Juice will deal 0 damage.

Screenshot from 2023-12-26 15-57-44

### DIFF
--- a/calc/src/mechanics/gen4.ts
+++ b/calc/src/mechanics/gen4.ts
@@ -88,7 +88,8 @@ export function calculateDPP(
     desc.moveBP = basePower;
   } else if (move.named('Judgment') && attacker.item && attacker.item.includes('Plate')) {
     move.type = getItemBoostType(attacker.item)!;
-  } else if (move.named('Natural Gift') && attacker.item && attacker.item.includes('Berry')) {
+  } else if (move.named('Natural Gift') && attacker.item && attacker.item.includes('Berry') &&
+      attacker.item !== 'Berry Juice') {
     const gift = getNaturalGift(gen, attacker.item)!;
     move.type = gift.t;
     move.bp = gift.p;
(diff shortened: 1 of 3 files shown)


shrianshChari:
Bumping.

ShivaD173:
This fails to take into account Berry Sweet, which also has the same wrong interaction. An easier approach than this I think would be to just change `items.ts:400` from 1 to 0bp.

shrianshChari:
> An easier approach than this I think would be to just change `items.ts:400` from 1 to 0bp.

The reason why I didn't want to do this was that the text that would display would be inconsistent with other items that don't interact with Natural Gift. Take an Abomasnow holding an Icy Rock:

Abomasnow Natural Gift vs. 168 HP Abomasnow: 0-0 (0 - 0%) -- possibly the worst move ever
image

To highlight that Berry Juice and Berry Sweet don't interact at all with Natural Gift, we want to ensure the text isn't different so that it isn't confusing to people who are familiar with how Natural Gift appears when using a berry.

This is what it would look like when changing the bp from 1 to 0:

Berry Juice Abomasnow Natural Gift (Normal) vs. 168 HP Abomasnow: 0-0 (0 - 0%) -- possibly the worst move ever
image

And this is what it would look like when explicitly telling Natural Gift to ignore Berry Juice and Berry Sweet:

Abomasnow Natural Gift vs. 168 HP Abomasnow: 0-0 (0 - 0%) -- possibly the worst move ever
image

As you can see, the second example is more in line with how Natural Gift interacts with other items, and I believe it is preferable to treat Berry Juice and Berry Sweet this way.

PR #596, opened 2024-01-11
Toxic fix

noobscrub27:
Fix several bugs:

- End of turn damage is now considered on the first turn.
- A pokemon that was KO'd by a move but would have healed above 0HP at the end of the turn is now considered KO'd.
- Toxic damage works now.
- Neutralizing Gas now negates end of turn healing and damage caused by abilities.
I also changed the way OHKOs are displayed when end of turn damage comes into play. For example, if a move isn't a guaranteed OHKO before end of turn damage, but has a higher chance to OHKO after end of turn damage, the chance to OHKO after end of turn damage is displayed separately. This is for cases where someone's pokemon outspeeds their opponent's pokemon, but their opponent can OHKO and the player wants to see how likely it is for them to KO their opponent before they get attacked. Here's an example: 0 SpA Ivysaur Weather Ball (100 BP Fire) vs. 0 HP / 252 SpD Abomasnow in Sun: 316-376 (98.4 - 117.1%) -- 87.5% chance to OHKO (guaranteed OHKO after poison damage)


noobscrub27:
Hi, thanks for the response!
In hindsight I could have done more to explain myself. By end of turn damage being considered on the first turn, I mean that when calculating damage, if the defender would lose health at the end the turn due to a status condition, it currently only starts counting this on the second turn.
As an example, this Mew is poisoned and seeded. It's level 30 with no HP IVs or EVs so it has exactly 100 HP. A level 99 Chansey's Seismic Toss should OHKO after hazards are accounted for, but the calculator thinks its a 2HKO.
image
And for toxic damage, not only is damage not applied on the first turn, the counter doesn't increase either.
Here's an example with a level 82 Chansey using Seismic Toss against a Pokemon with 200 Max HP:
image
This is listed as a 3HKO, but it's actually a 2HKO. This is how it should play out:

1. Seismic Toss: 200 - 82 = 118
2. Toxic damage (toxic count 1): 118 - 12 = 106
3. Seismic Toss: 106 - 82 = 24
4. Toxic damage (toxic count 2): 24 - 25 = -1 (2HKO)

This is how it the calculator does it currently:

1. Seismic Toss: 200 - 82 = 118
2. (EOT damage is skipped and the toxic count does not increase)
3. Seismic Toss: 118 - 82 = 36
4. Toxic damage (toxic count 1): 36 - 12 = 24
5. Seismic Toss: 24 - 82 = -58 (3HKO)

This is my first pull request so please be patient if I make any mistakes. Things are pretty busy in my life right now, but I'm going to do my best to fix this soon.

noobscrub27:
The rebase is finished. I added some new tests. The tests I added are for generations 3-9.
Here are the tests I wrote:
- KOed Pokemon don't receive HP recovery after 5+ turns
- KOed Pokemon don't receive HP recovery after 1-4 turns
- EoT damage is calculated correctly after 5+ turns
- EoT damage is calculated correctly after 1-4 turns
- EoT damage is calculated correctly on the first turn
The reason tests for 1-4HKOs are separate from tests for 5+HKOs is because the calculateKOChance function is used for the former and predictTotal is used for the latter. I also included some tests to ensure that in cases where a defender heals more at the end of the turn than they would lose (holding leftovers, receiving leech seed, etc.), they don't have their HP increased past 0 if the initial hit would have KOed.
Let me know if I should add any more tests, or remove some that I added. I put them all in the Multi-Gen section for now.
It's also worth noting that there are some issues with generation 1-2 EoT that my code doesn't address, namely the interactions between toxic and leech seed. This is something that I want to get around to at some point but I'm gonna be honest this has taken me a lot longer than I expected it to already so I'm gonna have to see it goes because early gen mechanics can be kind of a handful to work with. The calc handles toxic differently than it does other sources of EoT healing/damage, so I'm expecting having it work with leech seed would require the code for leech seed to be rewritten. This gets more complicated when you consider that the amount of healing a defender receives from the attacker's leech seed would also need to be changed.

noobscrub27:
No need to apologize. Thanks for all the help!

PR #599, opened 2024-02-06
"inputmode=decimal" for the numeric input text boxes

shrianshChari:
This is a change made mostly for mobile users - it will have the [decimal keyboard](https://css-tricks.com/everything-you-ever-wanted-to-know-about-inputmode/#aa-decimal) active when inputting the level, base stats, IVs, DVs, EVs (I'm pretty sure this was already the case), and the base power of moves.

### DIFF
--- a/src/honkalculate.template.html
+++ b/src/honkalculate.template.html
@@ -124,7 +124,7 @@
                         </div>
                         <div>
                             <label>Level</label>
-                            <input class="level" value="100" />
+                            <input class="level" value="100" inputmode="numeric" />
                         </div>
                         <div class="hide">
                             <label>Weight (kg)</label>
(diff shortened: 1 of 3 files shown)


monsanto:
my first thought is why you aren't using numeric instead of decimal, your commit message says it works better on ios but there is no elaboration or comment in the file itself

shrianshChari:
I had a link that would display how the virtual keyboard would look in my initial message, but essentially, on iOS, `decimal` will only pull up numbers on the virtual keyboard, whereas `numeric` would make it so that it makes the number keys very small and less useful.

From CSS-Tricks, this is what the `numeric` keyboard looks like on Android/iOS:

image

And this is what the `decimal` keyboard looks like:

image

PR #623, opened 2024-05-22
Helping Hand + Multi-Hit move fix

theastrogoth:
Helping Hand should only apply a boost to the first hit of a multi-hit move. I've only seen this tested in gen 9, but it's possible that this change could apply to previous gens as well.

Without HH:
      0 Atk Mamoswine Icicle Spear (3 hits) vs. 0 HP / 0 Def Hippowdon: 180-216 (50.4 - 60.5%) -- approx. 2HKO

Current with HH (boost across all 3 hits): 
      0 Atk Mamoswine Helping Hand Icicle Spear (3 hits) vs. 0 HP / 0 Def Hippowdon: 258-312 (72.2 - 87.3%) -- approx. 2HKO

 New with HH (only the first hit is boosted):
      0 Atk Mamoswine Helping Hand Icicle Spear (3 hits) vs. 0 HP / 0 Def Hippowdon: 206-248 (57.7 - 69.4%) -- approx. 2HKO

### DIFF
--- a/calc/src/mechanics/gen789.ts
+++ b/calc/src/mechanics/gen789.ts
@@ -631,6 +631,11 @@ export function calculateSMSSSV(
     const origDefBoost = desc.defenseBoost;
     const origAtkBoost = desc.attackBoost;
 
+    // Helping Hand only applies to the first hit of a multi-hit move
+    if (!(move.dropsStats && move.timesUsed! > 1)) {
+      field.attackerSide.isHelpingHand = false;
+    }
+
     let numAttacks = 1;
     if (move.dropsStats && move.timesUsed! > 1) {
       desc.moveTurns = `over ${move.timesUsed} turns`;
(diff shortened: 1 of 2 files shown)


DaWoblefet:
Do you have cartridge footage of this? When I [test that exact interaction](https://www.youtube.com/watch?v=G2OiHqMC3B8), all hits of Icicle Spear are boosted via Helping Hand. Mamoswine rolled 90 and 92 damage on its 2nd and 3rd hits, which are not possible damage rolls without Helping Hand. Moreover, we know from USUM that Helping Hand wears off during end-of-turn effects, not in checks done after move success (compare to e.g. resist Berries). You can do things like get multiple Helping Hand-boosted attacks via Helping Hand + an opponent using Instruct.

theastrogoth:
Thanks for running the test! It appears that the different behavior only applies to Tera Raids... oops. If I had to guess, the lack of clearly defined "turns" in raids led to Helping Hand being handled differently. 

Apologies for the mistake.

PR #629, opened 2024-07-12
feat: allow to Import showdown from GET parameters

arleypadua:
#### Overview
Recently I started writing a web-based version of a save editing app ([PKHeX Web](https://pkhex-web.github.io)). This change allows importing Pokemon data directly from PKHeX Web (or any other app) using a base64-encoded URL parameter named `import`

#### Changes Made

    - Decodes a base64-encoded string from the `import` URL parameter.
    - Sets the decoded string as the value of the text area for importing Pokemon data.

#### Purpose
Facilitates seamless integration between any external app and the Pokemon calculator. Allows users to transfer Pokemon data (party or box) from PKHeX Web to the calculator, streamlining their workflow.

### DIFF
--- a/src/js/index_randoms_controls.js
+++ b/src/js/index_randoms_controls.js
@@ -199,6 +199,17 @@ $(document).ready(function () {
 			}
 		}
 	}
+
+	let importParam = params.get('import');
+	if (importParam) {
+        try {
+            var decodedImport = atob(importParam); // Decode base64
+            $('.import-team-text').val(decodedImport); // Set value to text area
+        } catch (e) {
+            console.error('Failed to decode Import parameter:', e);
+        }
+	}
+	
 	$(".calc-trigger").bind("change keyup", PC_HANDLER);
 	performCalculations();
 });


AustinXII:
Thanks! Would you be able to fix the tests?

arleypadua:
> Thanks! Would you be able to fix the tests?

Thanks, I guess lint was complaining about the use of `let` as a unknown token.

Switched to use the plain old `var`, hope it fixes the issue

arleypadua:
> Hello, there are [still some build errors](https://github.com/smogon/damage-calc/actions/runs/10015969881/job/27689841815), but these should be easy to fix. You can run tests locally btw (`npm run test`).
> 
> This should be ready-to-merge once the build errors are fixed.

Whops, my bad.. as I'm on holidays, I thought that that was the only issue and ended up fixing on my phone.

Will find some time later this week to properly run it locally and fix everything ✌🏼

arleypadua:
Thanks again @AustinXII @thejetou 

Fixed all lint issues on the latest commit and tests are running green ✅ 

image

arleypadua:
Thanks a lot, don't know what happened since it fixed and passed locally.

Maybe I overlooked something. Thanks anyway

PR #635, opened 2024-07-26
Gen 8 Random Battles: Add support for Doubles and BDSP

shrianshChari:
Uses similar code to #632 but for Gen 8 instead of Gen 9

### DIFF
--- a/src/js/shared_controls.js
+++ b/src/js/shared_controls.js
@@ -623,8 +623,8 @@ $(".set-selector").change(function () {
 		var itemObj = pokeObj.find(".item");
 		var randset;
 		if ($("#randoms").prop("checked")) {
-			if (gen === 9) {
-				// The Gen 9 randdex contains information for multiple Random Battles formats for each Pokemon.
+			if (gen >= 8) {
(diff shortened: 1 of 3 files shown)


src/js/shared_controls.js
```diff
@@ -1280,6 +1280,36 @@ var SETDEX = [
...
+// An example using Absol would be:
+// {
+//		...
+//		Absol: {
+//			Randoms: {...},
+//			Doubles Randoms: {...},
+//			BDSP Randoms: {...}
+//		}
+//		...
+// }
```
Karthik99999, line 1300: I don't think you need this huge comment for every single gen

shrianshChari: I mostly just copied the gen 9 code and `sed`ed 8 to 9; I can remove it

AustinXII:
> At this point, we should just have a function that takes an object of format name -> sets then returns an object of pokemon -> format -> sets. This function then can be used for all gens that has multiple formats.

Good idea

shrianshChari:
> At this point, we should just have a function that takes an object of format name -> sets then returns an object of pokemon -> format -> sets. This function then can be used for all gens that has multiple formats.

Did

PR #637, opened 2024-07-30
Add "1 hit" option for multi-hit moves

lily-ac:
as discussed on discord https://discord.com/channels/630837856075513856/1058193416430825522/1267817692908355595

### DIFF
--- a/src/honkalculate.template.html
+++ b/src/honkalculate.template.html
@@ -493,6 +493,7 @@
                         <input class="move-stellar visually-hidden" type="checkbox" id="stellarL1" />
                         <label class="stellar-btn btn" style="display: none;" for="stellarL1" title="Make this attack Stellar boosted?">Stellar</label>
                         <select class="move-hits hide">
+                            <option value="1">1 hit</option>
                             <option value="2">2 hits</option>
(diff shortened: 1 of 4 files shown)


lily-ac:
thanks, is fixed
i agree tbh, i don't love this either, it's a janky workaround for functionality that was removed by a previous change

ideally the way damage numbers are displayed for multi-hit moves needs reworking altogether, because the current numbers are very misleading

236+ Atk Life Orb Shellder Icicle Spear (5 hits) vs. 36 HP / 0 Def Eviolite Tinkatink: 5-15 (22.7 - 68.1%) -- approx. 0.4% chance to 2HKO
Possible damage amounts: (5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 15)

this makes it look like you have a 1 in 16 chance to do 15 dmg, as opposed to the reality of 1 in 1,048,576
it also doesn't show that you can do 6 dmg for example, which would be a lot more likely to happen, and therefore helpful to know

i'm not sure what the solution looks like though, so if we can just have this workaround for now, that would be great :)

PR #642, opened 2024-09-14
Added browser icon

ColeCurrent:


ColeCurrent:
Its often hard to distinguish the damage calculator from other tabs with the default browser icon. When doing nuzlockes you often need to have multiple tabs of upcoming gyms, routes, or notes so you often have to distunguish tabs by browser icon.

PR #651, opened 2024-10-17
Fix Knock off Damage during Klutz/Magic Room

ShivaD173:
Add Disabled Item field that is used to check knock off damage during Klutz/Magic Room


AustinXII:
Thanks! Could you add a test for this please?

ShivaD173:
Added a Test, can this PR be merged?

PR #659, opened 2024-10-26
Jet Punch is no longer Sheer Force-boosted

shrianshChari:
Fixes https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/page-40#post-10317693


AustinXII:
Sorry about missing this!

PR #660, opened 2024-10-28
Add a box indicating the total number of EVs

shrianshChari:
image

Useful for EVing Pokemon and not exceeding the total number of EVs a Pokemon could have.


shrianshChari:
> I think this should be more in-line with the others total displays and show the number without a box around it.

Did, this is what it looks like now.

Screenshot from 2024-10-30 21-32-20

PR #667, opened 2024-11-03
Physical Shell Side Arm should make contact

shrianshChari:
Fixes https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/post-10315730


AustinXII:
Thanks!

PR #671, opened 2024-12-27
Fix EVs/IVs not getting used from randbats sets

Karthik99999:
The randbats sets use 3 letter stat names instead of the calc's regular 2 letter stat names, which caused stuff like min atk/spe and lowered hp evs to not get applied.


AustinXII:
Thanks!

Issue #675, opened 2025-01-06
Dual type damage calculation (Gen3)


will-jj:
When a Pokemon has two types in the Emerald damage calculator, the type effectiveness is carried out in two stages.  This can lead to differences of 1 HP between the ROM and the calculator due to truncation.

The damage calculator already has the [ordering](https://github.com/pret/pokeemerald/blob/50d325f081a161f4a223d999497d7a65bd896194/src/battle_main.c#L335) implemented, [here](https://github.com/smogon/damage-calc/blob/a5095a9aad8405ba311bd94e36fd89d792df1c50/calc/src/mechanics/gen3.ts#L76).
But multiplies the two together to calculate base damage, [here](https://github.com/smogon/damage-calc/blob/a5095a9aad8405ba311bd94e36fd89d792df1c50/calc/src/mechanics/gen3.ts#L117).  Differing from the Emerald source implementation, [here](https://github.com/pret/pokeemerald/blob/50d325f081a161f4a223d999497d7a65bd896194/src/battle_script_commands.c#L1355).

This has been observed and documented by Factory players (see image).
 
I have additionally made a small patch and tested against some potential cases in the Factory, it is noted that there was only the initial test case which was able to reproduce the issue - mainly included for interest (see table).
 
 For this case the critical calculation is as follows (simplification of Emerald code):
 
 ```c
 // TYPE_ELECTRIC, TYPE_ELECTRIC, TYPE_MUL_NOT_EFFECTIVE  (5)
 // 181 x 5 / 10 = 90
 damage = damage x 5 / 10 
 // TYPE_ELECTRIC, TYPE_FLYING, TYPE_MUL_SUPER_EFFECTIVE  (20)
 // 90 x 20 /10 = 180
 damage = damage x 20 /10
 ```
 
 Whereas in the damage calc:
 
 ```c
 // 0.5 x 2 = 1
 typeEffectiveness = type1Effectiveness * type2Effectiveness;
 // 181 x 1 = 181
 damage = damage * 1
 ```
 
 
 Finally, it is noted there is precedence for the application of type1 then type2 effectiveness within the damage calc with Gen4 already already implementing this, & Gen1/2 has in if statement such that it only applies to Gen 1, and not 2.
 
 ![image](https://github.com/user-attachments/assets/0ca6de87-06e2-4b1e-a502-e1621b652fcc)

[...]

PR #676, opened 2025-01-06
Apply type effectiveness in two stages to match implementation

will-jj:
This avoids any truncation in certain cases where the first type effectiveness is 0.5

Brings implementation inline with Gen1 & 4

Fixes: #675


will-jj:
Introduced multi-hit bug fixed by thejetou in:
https://github.com/smogon/damage-calc/commit/640d2d49df0431203fc16ca9c2082d8addd299ab

PR #680, opened 2025-01-20
Add sticky weather

will-jj:
This replicates the pattern (and code) of sticky move, and if a weather has been pinned then the auto set weather function will simply return early. 

Having sticky weather has been approved in #348 "Allow for 'pinning' weather".

### DIFF
--- a/src/css/dark-theme.css
+++ b/src/css/dark-theme.css
@@ -164,3 +164,11 @@ input.select2-input {
     */
     background: url('../js/vendor/select2/select2.png') no-repeat 100% -22px, var(--background);
 }
+
+.locked-move:checked + .btn {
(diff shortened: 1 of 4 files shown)


will-jj:
Note: I did not spend that long choosing the colour for the dark theme, you may want to suggest something more appropriate.

will-jj:
Great, thanks - it was still a bit of a WIP, with regards to the "last-clicked" - it could be perhaps set when it is changed/on load so it would feel a bit more natural.  For instance if you want to lock in none, & it is already on none, it will require three clicks. Or if you want to go back to none after an auto change, just one click will go back and lock it in.

It is definitely usable but that's mainly why I left it as a draft.

PR #704, opened 2025-06-09
No Trapping Damage for Gen 1

ShivaD173:


andrebastosdias:
Fixes https://github.com/smogon/damage-calc/issues/705

PR #709, opened 2025-07-20
Big root applies to OHKO

### LINKED ISSUE #677: Big Root does not update max health recovered
https://github.com/smogon/damage-calc/blob/e26ca06440ccb0d9f3a2929a9f7c90b9a6bce6c3/calc/src/desc.ts#L140

`max` is not updated along with `drained` when attacker has the "Big Root" item causing the displayed healed amount to be unchanged when the drain move is a OHKO.

GregLyons:
Fixes #677

First contribution, please let me know if anything else I should add to this PR

### DIFF
--- a/calc/src/desc.ts
+++ b/calc/src/desc.ts
@@ -149,13 +149,15 @@ export function getRecovery(
       [minD, maxD] = multiDamageRange(damage) as [number[], number[]];
     }
     const percentHealed = move.drain[0] / move.drain[1];
-    const max = Math.round(defender.curHP() * percentHealed);
+    const attackerHasBigRoot = (attacker.hasItem('Big Root'));
+    let maxDrain = Math.round(defender.curHP() * percentHealed);
+    if (attackerHasBigRoot) maxDrain = Math.trunc(maxDrain * 5324 / 4096);
     for (let i = 0; i < minD.length; i++) {
       const range = [minD[i], maxD[i]];
       for (const j in recovery) {
         let drained = Math.max(Math.round(range[j] * percentHealed), 1);
-        if (attacker.hasItem('Big Root')) drained = Math.trunc(drained * 5324 / 4096);
-        recovery[j] += Math.min(drained, max);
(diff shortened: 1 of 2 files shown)


calc/src/test/calc.test.ts
```diff
@@ -1180,6 +1180,22 @@ describe('calc', () => {
...
+        const weak = Pokemon('Abomasnow', {
+          item: 'Icy Rock',
+          ability: 'Snow Warning',
+          nature: 'Hasty',
+          evs: {atk: 252, spd: 4, spe: 252},
+          level: 29,
+        });
+        // Guaranteed OHKO
+        const result = calculate(bigRoot, weak, Move('Drain Punch'));
+        expect(result.range()).toEqual([120, 142]);
```
GregLyons, line 1195: Got from existing calc

calc/src/test/calc.test.ts
```diff
@@ -1180,6 +1180,22 @@ describe('calc', () => {
...
+        // 100 damage * (50% heal * ~1.3 big root boost = ~64.99% heal) truncates to 64 HP recovered
+        expect(result.recovery().recovery).toEqual([64, 64]);
```
GregLyons, line 1197: Without this change, this returns `[50, 50]`

calc/src/desc.ts
```diff
@@ -149,13 +149,15 @@ export function getRecovery(
       [minD, maxD] = multiDamageRange(damage) as [number[], number[]];
     }
     const percentHealed = move.drain[0] / move.drain[1];
-    const max = Math.round(defender.curHP() * percentHealed);
+    const attackerHasBigRoot = (attacker.hasItem('Big Root'));
+    let maxDrain = Math.round(defender.curHP() * percentHealed);
+    if (attackerHasBigRoot) maxDrain = Math.trunc(maxDrain * 5324 / 4096);
```
GregLyons, line 154: To be honest, I'm not entirely sure whether this big root factor needs to take place before or after rounding; am going based on the calculation for `drained`

PR #732, opened 2025-10-17
Added Legends Z-A Megas to damage calc

Daniel1464:
For now, they use the placeholder abilities(like showdown does).


thejetou:
We can `.skip` the failing tests but we should alphabetically order the items & Pokemon.

PR #733, opened 2025-10-17
Added new Z-A move(Nihil Light)

Daniel1464:
It's a 200(changed to 100) BP move that ignores fairy type and stat changes.

For now, the move is mostly a placeholder since we don't have data on it from champions


thejetou:
Thanks!

PR #744, opened 2025-11-23
Fix+Enhance Wonder Room and Power Trick handling

Celestia74:
wonder room swaps stats, then applies stage changes. currently this is behaving in the opposite manner

### DIFF
--- a/calc/src/mechanics/gen789.ts
+++ b/calc/src/mechanics/gen789.ts
@@ -106,11 +106,6 @@ export function calculateSMSSSV(
     moveName: move.name,
     defenderName: defender.name,
     isDefenderDynamaxed: defender.isDynamaxed,
-    isPowerTrickAttacker: (
-      (move.named('Foul Play') || (move.category === 'Special' && !move.named('Body Press')))
-        ? false
-        : field.attackerSide.isPowerTrick
-    ),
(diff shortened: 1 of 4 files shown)


Celestia74:
did some additional research into what was going on

in essentia I'd be reverting 999b0fb
The intent there was to make Body Press work correctly with its weird using SpD in Wonder Room quirk, but it ended up making _everything_ use the wrong stat. All this is currently doing is reverting that.
Will be making another commit on this branch to add proper exceptions to Body Press and Shell Side Arm.

calc/src/mechanics/gen789.ts
```diff
@@ -670,8 +670,14 @@ export function calculateSMSSSV(
   }
   result.damage = childDamage ? [damage, childDamage] : damage;
 
-  desc.attackBoost =
-    move.named('Foul Play') ? defender.boosts[attackStat] : attacker.boosts[attackStat];
+  if (move.named('Foul Play')) {
+  	desc.attackBoost = defender.boosts[attackStat];
+//this feels hacky but the coder who made it couldn't think of a better way
```
thejetou: I don't really feel like it's hacky to be honest.

thejetou (commented):
I think Wonder Room is now finicky enough that it needs its own tests.

PR #782, opened 2026-04-03
updated a few move type overrides

Celestia74:
- Revelation Dance follows secondary type if primary type is null.
- Kantonian Tauros is locked to normal Raging Bull regardless of its own type or Raging Bull's base type.
- same with Ogerpon-Teal and grass Ivy Cudgel
- Morpeko-Hangry is locked to dark Aura Wheel, but no lock exists for normal forme.


thejetou:
Thank you.

PR #787, opened 2026-04-17
simplify Champions dex as a whitelist and changelog

Celestia74:
Instead of having to redeclare every Pokemon and move, import data from latest gen and make changes as needed. Saves a ton of space and makes it easier to add and remove things as rulesets change.


thejetou:
Thanks!

PR #788, opened 2026-04-18
Champions: Paralysis should halve speed

shrianshChari:
Fixes https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/post-10958741


thejetou:
Thanks!

Issue #756, opened 2026-02-28
Layout breaks when font scaling is set to Large


Denhai:
The setting is in chrome://settings/appearance and it can be change to Large or Very Large

<img width="1547" height="941" alt="Image" src="https://github.com/user-attachments/assets/c4ae9825-615b-4bdb-ac34-6e29512a3ba9" />

This appears to be caused by `em`-based widths on inputs.

Specifically these inputs have a larger than normal font-size, and in turn scaled width:
https://github.com/smogon/damage-calc/blob/9b1ab6c1941715875eb0ca530fcc6f0d8a0e5c1f/src/index.html#L1076-L1077

Switching to `rem` units fixed it for me, I'm not sure if that's the right fix though.

PR #790, opened 2026-04-20
Fixed Issue #756: Layout breaks when font scaling is set to Large

a22303324HugoPereira:
From what I can tell changing the em to rem seemed to be enough, I assume they didn't want the actual letter size changed and just the boxes so they wouldn't deform the layout of the page.


thejetou:
Thanks!

PR #792, opened 2026-04-23
Champions: Fix and test Mega Sol

shrianshChari:
https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/page-45#post-10961435


thejetou:
Thanks!

Issue #139, opened 2017-03-29
Add option for not automatically assuming max EVs in RBY/GSC sections


SadisticMystic:
Obviously in a competitive battle under those two rulesets, it's possible to max out all the stats so for almost all practical purposes you might as well. However, there are some cases (the most prominent one on my mind is setting up in-game scenarios) where more flexibility in the gamut of player/opponent stats is helpful.

A lot of the time, it's possible to construct the equivalent setup in ADV mode and run the numbers from there, but there are some details that don't quite sync across the generation gap (in particular, the damage variation roll dropping from a d39 to a d16) for cases where the slight changes might tangibly affect the odds.

Having the EV text boxes be hidden by default, and only revealing them on the deselection of a toggle switch like the one used for critical hits, would be fine.

AustinXII:
I don't know if we should cater to the needs of ingame calculation as this calc is designed specifically for Pokémon Showdown. It has the flexibility to be used for other purposes with a little extra manual effort which is fine imo. There's also a concern on what the evs should be set to as they are automatically set from the ``setdex`` files. Regardless in the end many people will still have to manually change evs

AustinXII:
Oh I misunderstood, I suppose some
Players don't have to max out all their EVs or may have not done so by accident, this will be fixed soon(tm)

AustinXII:
(If anyone else wants to do this I can explain exactly what's needed to be done in order to fix this, I'm just busy with life atm)

AustinXII:
Hmm I did work on this and almost finished at one point (I should really start saving my work even if it’s abandoned) changing the HTML is pretty easy I believe changing ``shared_controls.js`` is that hard part. If you need a step in the right direction I’ll help

AustinXII:
@Ntonio36

Ntonio36:
HTML is easy, I can relate, but I need to know what am I exactly supposed to do for `shared_controls.js`... or let's say, what does @SadisticMystic exactly want to do

AustinXII:
@Ntonio36 Once you have the HTML the next step isn’t that hard it’s just the hard part. Just allow the EV editor to be customizable in gens 1 and 2. You can see how EVs are customizable in other gens for reference.

Ntonio36:
Alright, I have now to look for it and see how it can be made. Will need 10 days at most. I guess.

Ntonio36:
[...]

PR #797, opened 2026-05-05
Add support for EVs and importing DVs within RBY and GSC

### LINKED ISSUE #664: Custom DVs ignored
### Problem

In the gens that use DVs, DV spreads may only be either 15 in everything or a default hidden power spread. Custom spreads can be achieved by editing after selection, but they can not appear initially.

The only built-in set where this is relevant is the gen 1 Mewtwo "Ubers Staller" set, which should have 13 special DVs (to avoid overflow when boosted to +4) but gets 15 instead. 

Sets with hidden power get assigned 15 DVs in everything, then get their DVs corrected to give them a default spread. For example, the following custom set gets assigned 13 attack DVs in gen 2:
> Abra
> IVs: 2 Atk
> \- Hidden Power Steel

### Cause

The assignment to 15 DVs happens [here](https://github.com/smogon/damage-calc/blob/9e8ca8eaa74f9d5397048feb490a6c54d0301956/src/js/shared_controls.js#L685) and the correction to hidden power default spread happens [here](https://github.com/smogon/damage-calc/blob/9e8ca8eaa74f9d5397048feb490a6c54d0301956/src/js/shared_controls.js#L488).

The overwrite to max DVs occurs because built-in sets don't have a "dvs" property & custom sets aren't given one on import. The correction to default hidden power spreads happens because of the overwrite.

### Solution

I think the best solution would be to remove all references to the "dvs" property in setdex entries and, instead, always calculate DVs from IVs. As far as I can tell, the only place that assigns a dvs entry is [correctHiddenPower](https://github.com/smogon/damage-calc/blob/9e8ca8eaa74f9d5397048feb490a6c54d0301956/src/js/shared_controls.js#L985) and it's only read [here](https://github.com/smogon/damage-calc/blob/83d3831d19a60c01baa8e6bfa82653593e40661d/src/js/shared_controls.js#L685) and [here](https://github.com/smogon/damage-calc/blob/83d3831d19a60c01baa8e6bfa82653593e40661d/src/js/shared_controls.js#L992).

If the dvs property needs to be kept, another solution would be to change
https://github.com/smogon/damage-calc/blob/9e8ca8eaa74f9d5397048feb490a6c54d0301956/src/js/shared_controls.js#L685
to something like
`set.dvs?.[LEGACY_STATS[gen][1]] ?? calc.Stats.IVToDV(set.ivs?.[LEGACY_STATS[gen][0]] ?? 31);`

Either way I _think_ the [IV/DV correction code](https://github.com/smogon/damage-calc/blob/9e8ca8eaa74f9d5397048feb490a6c54d0301956/src/js/shared_controls.js#L477) can be deleted. I don't think it's ever doing anything unless other code messes things up. If it stays, maybe call correctHiddenPower instead of re-implementing its behaviour.

Tree69420:
Fixes #664 and #139 by adding similar frontend options as in the other tiers, adding IV checks for gen 1 and 2 in the moveset importer, and changing the gen 1 stat calculator to use EVs


thejetou:
The build is failing and also your branch seems to include unrelated changes.

Tree69420:
oops i had my commit on a different repo that wasnt considered a fork so i had to do some wack shit to merge it and i think thats the problem

package.json
```diff
@@ -5,7 +5,7 @@
   "license": "MIT",
   "devDependencies": {
     "eslint": "^8.57.0",
-    "express": "^4.22.1",
+    "express": "^4.21.2",
```
Daniel1464: Why downgrade this dependecy?

Daniel1464 (changes requested):

package.json
Tree69420: definitely an automatic thing when i merged up with my most recent pull mb

src/js/moveset_import.js
```diff
@@ -102,6 +102,22 @@ function statToLegacyStat(stat) {
 	}
 }
 
+
+function statToLegacyStatGen1(stat) {
```
thejetou: I think it reads better to just change `statToLegacyStat` to accept a `gen` param then returning `sl` or `sa` based on the gen.

src/js/moveset_import.js
```diff
@@ -152,20 +169,38 @@ function getStats(currentPoke, rows, x) {
 			currentPoke.level = parseInt(currentRow[1].trim());
 			break;
 		case 'EVs':
+			if (gen === 1) {
```
thejetou, line 159: Why do we zero the evs?

src/js/moveset_import.js
```diff
@@ -152,20 +169,38 @@ function getStats(currentPoke, rows, x) {
...
 			currentPoke[$('#champions').prop('checked') ? 'sps' : 'evs'] = evs;
 			break;
 		case 'IVs':
 			for (j = 1; j < currentRow.length; j++) {
 				currentIV = currentRow[j].trim().split(" ");
-				currentIV[1] = statToLegacyStat(currentIV[1].toLowerCase());
-				ivs[currentIV[1]] = parseInt(currentIV[0]);
+				if (gen === 1) currentIV[1] = statToLegacyStatGen1(currentIV[1].toLowerCase());
+				else currentIV[1] = statToLegacyStat(currentIV[1].toLowerCase());
+				if (gen === 1 || gen === 2) dvs[currentIV[1]] = parseInt(currentIV[0]);
```
thejetou: PS uses 0 - 30 for DVs so we need to divide by 2.

src/js/shared_controls.js
```diff
@@ -676,14 +676,19 @@ $(".set-selector").change(function () {
 			pokeObj.find(".level").val(set.level === undefined ? 100 : set.level);
 			for (i = 0; i < LEGACY_STATS[gen].length; i++) {
 				var stat = $("#randoms").prop("checked") ? legacyStatToStat(LEGACY_STATS[gen][i]) : LEGACY_STATS[gen][i];
-				if ($("#champions").prop("checked") && !set.sps) {
-					var sps = set.evs && set.evs[stat] !== undefined ? set.evs[stat] : 0;
-					if (sps === 4) sps = 1;
-					else sps = Math.ceil(sps / 8);
-					pokeObj.find("." + LEGACY_STATS[gen][i] + " .sps").val(sps);
+				if (gen > 0 && gen < 3) {
```
thejetou: Simpler to remove this if condition and instead do something like:
```js
var evsDefault = gen > 2 ? 0 : $("#randoms").prop("checked") ? 84 : 252;
pokeObj.find("." + LEGACY_STATS[gen][i] + " .evs").val(
					(set.evs && set.evs[stat] !== undefined) ? set.evs[stat] : evsDefault);
```
(Define `evsDefault` outside of the loop though.)

src/js/moveset_import.js
Tree69420: On PS set exporting, if a pokemon has no EVs in a stat the stat just won't be included on the set export, so we can't manually zero them when they appear

src/js/moveset_import.js
```diff
@@ -152,20 +156,36 @@ function getStats(currentPoke, rows, x) {
...
 			}
 			currentPoke[$('#champions').prop('checked') ? 'sps' : 'evs'] = evs;
 			break;
 		case 'IVs':
 			for (j = 1; j < currentRow.length; j++) {
 				currentIV = currentRow[j].trim().split(" ");
-				currentIV[1] = statToLegacyStat(currentIV[1].toLowerCase());
-				ivs[currentIV[1]] = parseInt(currentIV[0]);
+				currentIV[1] = statToLegacyStat(currentIV[1].toLowerCase(), gen);
+				if (gen === 1 || gen === 2) dvs[currentIV[1]] = parseInt(currentIV[0]) / 2;
```
thejetou:
```suggestion
				if (gen === 1 || gen === 2) dvs[currentIV[1]] = Math.floor(parseInt(currentIV[0]) / 2);
```

src/js/shared_controls.js
```diff
@@ -674,16 +674,19 @@ $(".set-selector").change(function () {
 				pokeObj.find(".teraType").val(set.teraType || getForcedTeraType(pokemonName) || pokemon.types[0]);
 			}
 			pokeObj.find(".level").val(set.level === undefined ? 100 : set.level);
+			var evsDefault = gen < 3 ? 252 : ($("#randoms").prop("checked") ? 84 : 0);
+			pokeObj.find("." + LEGACY_STATS[gen][i] + " .evs").val(
+				(set.evs && set.evs[stat] !== undefined) ? set.evs[stat] : evsDefault);
```
thejetou:
```suggestion
```

thejetou (approved):
Will merge on Monday.

thejetou:
Thank you!

PR #801, opened 2026-05-10
Champions: Fix exported info missing Nature

Suyooo:
Hi! I noticed that exporting sets from the Champions calculator didn't include the Pokemon's nature, so here's a quick and small PR to fix that.


thejetou:
Thanks!

PR #804, opened 2026-05-11
Fix Last Respects base power scaling

FakedPlains:
## Summary
- Add Last Respects base power scaling from fainted allies, capped at 200 BP.
- Cover both Champions and Gen 9 damage calculation paths.
- Add regression coverage for normal scaling and the cap.

## Test plan
- npm test -- --runInBand

Made with [Cursor](https://cursor.com)


thejetou:
Can you add a UI for this?

Celestia74:
Why is this being capped to 200?
Singles sees it go to 350 in a single standard game (5 allies fainted), and the hardcoded limit is 5050BP.

Lusamine:
I assume the 200 is for VGC where you're likely to get 3 fainted allies, but still this should accommodate different modes.

FakedPlains:
> 为什么上限是 200？ 单打比赛中，一场标准比赛（5 个队友倒下）最高可达 350，而硬编码的上限是 5050BP。

This only considers the situation of VGC matches, I will modify the settings to a hard-coded upper limit of 5050 BP.

FakedPlains:
> Can you add a UI for this?

I've modified the moves selection area below. If you choose “Last Respects”, the “alliesFainted” will be displayed in the background, similar to the display method for selecting the “Kingambit” with “Supreme Overlord” ability.

PR #808, opened 2026-06-05
Fix Growth type in Champions

dyp1109:
Growth in champions has changed to grass type move. I have updated the champions_patch to include the changes.
[https://www.serebii.net/pokemonchampions/updatedattacks.shtml](url)


Celestia74:
It was in fact changed in Z-A, not Champions.
Z-A, however, is currently being bundled in with gen 9 due to the popularity of the Z-A OU mod. In gen 9, Growth is still normal type.
Growth's change is already documented in ZA_PATCH and has been commented out until/if the time comes to move the Z-A megas forwards.

dyp1109:
Ok, thank you very much for the clarification, I will close this PR now.

PR #810, opened 2026-06-07
Randoms: Fix Gen 8 and Gen 8 Doubles

shrianshChari:
Fixes https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/post-11019710


thejetou:
Thanks.

PR #813, opened 2026-06-17
Update Champions for M-B

Celestia74:
also added the rest of the known move changes just so they're already there for the future


flipperbw:
will this PR also be responsible for adding the logic for Life Orb/Muscle Band/Expert Belt (etc.) into the damage calculations?

Celestia74:
...yes it will, I just forgot. please hold

thejetou:
Great work, thank you!

PR #815, opened 2026-06-18
Champions: Update abilities

Celestia74:
When looking to just list legally usable abilities, we got a good few wrong.


thejetou:
Thank you.

PR #827, opened 2026-07-05
Support Nightmare

shrianshChari:
For https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/post-11047939


thejetou:
Thank you!

PR #831, opened 2026-07-08
fix Champions stat boosts

Celestia74:
fixes https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/post-11053363


thejetou:
Thanks.

PR #837, opened 2026-07-19
Support Charge

shrianshChari:
Fixes https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/post-11035239

### DIFF
--- a/calc/src/mechanics/gen3.ts
+++ b/calc/src/mechanics/gen3.ts
@@ -148,7 +148,7 @@ export function calculateADV(
   if (bp === 0) {
     return result;
   }
-  bp = calculateBPModsADV(attacker, move, desc, bp);
+  bp = calculateBPModsADV(attacker, move, field, desc, bp);
 
   const isCritical = move.isCrit && !defender.hasAbility('Battle Armor', 'Shell Armor');
   const at = calculateAttackADV(gen, attacker, defender, move, desc, isCritical);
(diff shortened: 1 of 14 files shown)


calc/src/mechanics/gen3.ts
```diff
@@ -269,6 +270,12 @@ export function calculateBPModsADV(
     basePower = Math.floor(basePower * 1.5);
     desc.attackerAbility = attacker.ability;
   }
+
+  if (field.attackerSide.isCharge && move.hasType('Electric')) {
```
thejetou: Charge is not a BP mod in Gen 3. It multiplies by 2 in the last steps after the `+ 2` along with critical hit, Helping Hand calculations etc.

calc/src/mechanics/gen4.ts
```diff
@@ -466,6 +466,11 @@ export function calculateBPModsDPP(
     desc.attackerAbility = attacker.ability;
   }
 
+  if (field.attackerSide.isCharge && move.hasType('Electric')) {
```
thejetou, line 464: Charge is calculated before Technician like Helping Hand.

calc/src/mechanics/gen3.ts
```diff
@@ -257,6 +257,7 @@ export function calculateBasePowerADV(
 export function calculateBPModsADV(
   attacker: Pokemon,
   move: Move,
+  field: Field,
```
thejetou: unused param

thejetou:
Thanks!

PR #839, opened 2026-08-04
Fix Dragonize and adjust some syntax

Celestia74:
https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/post-11077061


thejetou:
Thanks.

PR #843, opened 2026-08-13
Champions: Battle Armor should block critical hits

### LINKED ISSUE #841: Champions.ts - Battle Armor - Not Working
src/mechanics/champions.ts
const isCritical covers "Shell Armor" - works as expected
0 Atk Meowscarada Flower Trick vs. 0 HP / 0 Def Samurott: 102-120 (60 - 70.5%) -- guaranteed 2HKO

However, "Battle Armor" is not listed - same effect
const isCritical does not cover "Battle Armor" - does not work as expected
0 Atk Meowscarada Flower Trick vs. 0 HP / 0 Def Falinks on a **critical** hit: 66-78 (47.1 - 55.7%) -- 67.2% chance to 2HKO
Flower Trick is not expected to Critical Strike against Pokemon with Battle Armor

shrianshChari:
Fixes #841

### DIFF
--- a/calc/src/mechanics/champions.ts
+++ b/calc/src/mechanics/champions.ts
@@ -128,7 +128,7 @@ export function calculateChampions(
 
   // Merciless does not ignore Shell Armor, damage dealt to a poisoned Pokemon with Shell Armor
   // will not be a critical hit (UltiMario)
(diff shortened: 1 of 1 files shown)


thejetou:
Thanks!

PR #848, opened 2026-08-19
Stealth Rock should deal minimum 1 HP

shrianshChari:
Fixes https://www.smogon.com/forums/threads/pok%C3%A9mon-showdown-damage-calculator.3593546/post-11100290


thejetou:
Thanks!

PR #853, opened 2026-08-23
Guard the Knock Off mega stone item lookup

### LINKED ISSUE #852: Champions: calculate() throws when defender's item is not in the gen-0 item table
Repro at master `636e5b9`:

```js
const gen = Generations.get(0);
calculate(gen,
  new Pokemon(gen, 'Kingambit'),
  new Pokemon(gen, 'Incineroar', {item: 'Black Sludge'}),
  new Move(gen, 'Iron Head'),
  new Field());
// TypeError: Cannot read properties of undefined (reading 'megaStone')
//     at calculateBPModsChampions
```

Any move triggers it — the `resistedKnockOffDamage` pre-computation runs unconditionally and dereferences `gen.items.get(toID(defenderItem))` unguarded, which is `undefined` for ids outside the curated gen-0 pool (Black Sludge, Sticky Barb, …).

Fix: `item?.megaStone` (unknown item ⇒ not a mega stone). Same unguarded pattern in `calculateBPModsSMSSSV` (`gen789.ts`), reachable only if an id misses that gen's table.

comstock-jackson:
Fixes #852.

`calculateBPModsChampions`, `calculateBPModsSMSSSV` and `calculateBPModsBWXY` compute `resistedKnockOffDamage` for every standard damaging move, dereferencing `gen.items.get(toID(defenderItem))!`. The lookup returns `undefined` for any held item outside that gen's table, so `calculate()` throws. Trivially reachable in Champions, whose curated item table omits items valid in other gens (e.g. Black Sludge). Reachable in gens 5-9 only if an id misses those tables.

Fix: drop the non-null assertion and guard with `item?.megaStone`. An unknown held item is not a mega stone and stays knock-off-able. Regression test added in the Champions block.


thejetou:
Thanks.

Issue #745, opened 2025-11-24
Number of hits to KO doesn't account for consuming damage-resisting berries


Pinsplash:
<img width="1314" height="978" alt="Image" src="https://github.com/user-attachments/assets/45cc9392-a1dd-422c-9bdf-2d5fcf3847cb" />

The damage calculation itself is right, but this is really a guaranteed 2hko unless the target somehow had a second resist berry put onto it. If you told me this was a 2hko bordering on 3hko, I would expect that to mean the first hit does 33% and the 2nd hit does 66%.

Celestia74:
Additional information for other developers, from someone who wasn't smart enough to know how to fix it themself:

Currently, move.hits and move.timesUsed use the same code (which is why they can't be both enabled at once), which returns a lump sum of all hits. This runs into an issue when calculating end-of-turn effects, since for the purposes of computeKOChance all [x] timesUsed are being calculated as a single use of [x] hits.

To fix this, move.timesUsed needs to be separated from move.hits in some way to allow for EoT in between. The problem I couldn't figure out was how to allow it to continue calculations on further hits when all EoT is handled in desc.ts, _after_ the damage has already been calculated.

dyp1109:
I will share my findings here, 

<img width="1367" height="744" alt="Image" src="https://github.com/user-attachments/assets/7eb36a8f-13c4-481f-a2f4-51510f6ede91" />

There's a dropdown box below each move (bottom right), when "once" is chosen, the damage calculation will assumes the condition doesn't change (e.g. Colbur Berry does not consumed), so the 2HKO calculation is incorrect.

<img width="1325" height="854" alt="Image" src="https://github.com/user-attachments/assets/241905c3-d706-45b8-8993-43bd099bc7e6" />

However, if "Twice" is selected in the dropdown box, the item consumed is considered in the 2nd turn calculation, as you can see in
```
Possible damage amounts: (1st Hit: 73, 74, 75, 76, 77, 78, 79, 80, 80, 81, 82, 83, 84, 85, 86, 87)
(2nd Hit: 146, 148, 150, 152, 154, 156, 158, 160, 160, 162, 164, 166, 168, 170, 172, 174)
```
Stat change is also considered in that situation, for example

<img width="1299" height="754" alt="Image" src="https://github.com/user-attachments/assets/1503df59-8570-48fd-ba90-265fb3b5421c" />

[...]

PR #854, opened 2026-08-23
Gen 3/4: fix timesUsed condition so all damaging moves replay across multiple turns.

dyp1109:
Related issue: #745 

Gen 3–4 only replayed a move across turns when `dropsStats` was set (Overheat, Close Combat, etc.). Picking “Twice” on something like Sucker Punch still set `timesUsed > 1`, but the extra hits were never calculated, so Colbur Berry stayed on and the KO line became “not a KO”.

Gen 5+ already replays every damaging move. This removes the same `dropsStats` gate in Gen 3–4 so item consumption and mid-sequence stat updates apply on later turns (e.g. Sucker Punch vs. Colbur Berry over 2 turns).


thejetou:
Thanks.

PR #855, opened 2026-09-01
Add newly-revealed mega evolution abilities

Daniel1464:
Sharpness for mega absol Z, Aura Guard for mega lucario Z and Levitate for garchomp z.


thejetou:
Thank you.

